PR #39, opened 2015-03-11
Update README.md

shellygill:
This sample no longer uses Callout, so I have removed that from the readme.


doneill:
@shellygill thanks for the submission, looks good.

PR #60, opened 2015-07-01
Use of find method to get the location using the List of LocationSuggestionResult

mani8177:
@doneill This PR does the following 

1. In the `suggest` method, the location of the suggestions are fetched and stored in a `Map<Address,Location>` using the `Locator.find`.
2. When a suggestion is clicked, the address is used to lookup in the map and fetch the Location.
3. Lastly, the result is displayed on the map.

This solves the issue **Runtime | PlaceSearch Sample is not displaying address correctly**


mani8177:
@doneill I have refactored the PR as per your review comments.

doneill:
@mani8177 In the ```locatorParams()``` method, the ```String type``` parameter is always ```SUGGEST_PLACE```.  Can we remove the parameter and constant?

mani8177:
@doneill On it. The constant was part of implementing suggestions as we were using both the `LocatorSuggestionParameters` and `LocatorFindParameters`. I will remove the unnecessary parameters and constant

mani8177:
@doneill The Changes are done

doneill:
@mani8177 Looks good, will merge.

PR #64, opened 2015-07-07
Performance and search on submit

mani8177:
@doneill I have made the following changes 

1. Instead of saving the Location and it's address in the map as key-value pair and fetching it when an item is clicked, stored the `LocatorSuggestionResult` in an arrayList. Item is fetched from the ArrayList when a suggestion is clicked and `find` method is used in `AsyncTask` to get the location and display it.

2. Implemented `LocatorAsyncTask` to search when query is submitted/entered. 

3. Handled **Place Search** with both the functionalities 
      1. Selecting an item from Suggestion List
      2. Hitting enter after typing in a query
using boolean flags because the result address is displayed in the search textbox. After Geocoding, the result is displayed on the map and in the textbox, the method `onQueryTextSubmit` would each time pick this query as a new query and would go in a loop.

### DIFF
--- a/PlaceSearch/src/main/java/com/arcgis/android/samples/search/placesearch/MainActivity.java
+++ b/PlaceSearch/src/main/java/com/arcgis/android/samples/search/placesearch/MainActivity.java
@@ -197,6 +197,12 @@ public class MainActivity extends Activity {
     mSearchView.setIconifiedByDefault(false);
     mSearchView.setQueryHint(getResources().getString(R.string.search_hint));
 
+    // Open the soft keyboard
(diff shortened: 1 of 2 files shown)


doneill:
@mani8177 Update looks good, nice work.  Beyond the minor comments I have can you ensure the keyboard opens when you tap the search button on initial start of the app?  Currently you have to tap in the text box to get the keyboard to show which is 2 taps to start typing.

mani8177:
@doneill accommodated the review comments

doneill:
@mani8177 Updates look good, will merge.

PR #69, opened 2015-07-27
Popup for editing in OfflineEditor

mani8177:
@willcrick @doneill  Please review the update. Following changes are done as per your feedback on last PR:
1. Callout is displayed when offline
2. Tap on the callout to edit the attributes in a popup
3. After editing/adding a feature and clicking on save, the mid-points and end-points are not displayed.
4. Updated the README.md and added workflow to edit the attributes in a popup.

Do let me know your review comments if something should be changed/added.


doneill:
@mani8177 Updates look good and workflow works as well.  I have some additional comments to the inline comments provided:

**OfflineEditorActivity**
- Unnecessary semicolon ';' line 746
- Call ```toArray()``` with zero length array argument ```new Feature[0]```, replace with correctly sized array ```return features.toArray(new Feature[features.size()]);``` lines 235 & 883
- Remove redundant initializer ```new TreeMap()``` line 257
- Remove variable ```temp``` which is never used line 275
- Remove variable ```envelope``` which is never used line 558
- Remove variable ```gdbFeatureSelectedId``` which is never used at line 750
- Remove redundant initializer ```new TreeMap()``` line 909

**PopupForEditFeatureLayer**
- ```FILL_PARENT``` is deprecated, replace with ```MATCH_PARENT``` lines 519 & 520
- Call ```toArray()``` with zero length array argument ```new Feature[0]```, replace with correctly sized array ```return features.toArray(new Feature[features.size()]);``` line 150
- Remove variable ```env``` which is never used line 90

mani8177:
@doneill I have refactored the sample to use GeodatabaseFeatureServiceTable when adding the layers (Will suggested to use this method), also have accommodated the comments. 
The code had `RunQueryFeatureLayerTask` and `RunQueryLocalFeatureLayerTask`, two similar classes only difference was the callout being called. Now we have only one class hence avoiding the repetition of code. Also, Will suggested to show same callout with details when online/offline to maintain the look and feel. Lastly, Features are very quickly fetched and the spinner was displayed and gone in a flash and wouldn't be a good user experience hence removed it. 

Can you please review the changes?

PR #71, opened 2015-07-27
The default date was hardcoded to December 2011. I made it so the cur…

eric906:
Now, the current date will be used as the default in the Data Picker. Before, the default date was hard-coded to be December 2011.


doneill:
@eric906 Updates look good, I added a few fixes to #72 so will close this PR

Issue #78, opened 2015-09-06
Quartz Samples: Draw mode icon out of sync


n8ebel:
While drawing, if user taps draw mode icon again (rather than long press) the drawing finishes, but when they tap the icon again, it doesn't enter selected state and is unclear if drawing mode was activated.

Repro steps:

1. tap point icon to enter point drawing mode
2. tap to draw point
3. tap point icon again.  The drawing will finish correctly
4. tap point icon again to enter point drawing mode.  The icon will not be selected, but user can still draw

doneill:
@n8ebel Good Catch!  Thanks for creating this issue and creating a PR #79, I will review.

doneill:
Verified fixed

PR #79, opened 2015-09-06
bug: Fixed bug with draw mode icon being out of sync

n8ebel:
When drawing, if the user clicked the icon again (rather than long press) the drawing would complete, but then the next time the icon was pressed it would not enter the 'selected' state and would therefore be unclear to the user whether they were actually able to draw or not.

This was addressed by setting the draw mode to 'NONE' explicitly when unselected a draw mode icon

#78


doneill:
@n8ebel Fix looks good, I do have a comment for you to review.

PR #90, opened 2015-10-02
Dop/optimize modules

doneill:


doneill:
Prep for 10.2.7 release

PR #98, opened 2015-11-03
Dop/optimize samples

doneill:


doneill:
Some optimizations to current samples

PR #101, opened 2015-11-30
Dop/analyze samples

doneill:
Updated Nearby sample


doneill:
@shellygill Can you review these updates to refactor the API and use support lib for backward compatibility with older devices?

PR #139, opened 2016-04-13
better fix for ensuring that the view is not null

PuneetEsri:


doneill:
@PuneetEsri Looks good!

PR #140, opened 2016-04-14
Snodnipper osm layer demo

doneill:
Added some minor updates to the sample


doneill:
@snodnipper I reviewed and accepted for our sample. I added some minor updates.  Thanks and keep them coming :-).

PR #141, opened 2016-04-16
Updated to graphic renderer sample

doneill:


doneill:
@shellygill Please review these updates.  Removed the dependency on the `MapView` listener.  In sync with common sample design.

add-graphics-renderer/build.gradle
```diff
@@ -20,5 +20,5 @@ android {
 }
 
 dependencies {
-    compile 'com.android.support:appcompat-v7:23.2.1'
+    compile 'com.android.support:appcompat-v7:23.3.0'
```
shellygill: Would be nice to use a consistent version - should we update all to the newer version?

add-graphics-renderer/src/main/java/com/esri/arcgisruntime/sample/addgraphicsrenderer/MainActivity.java
```diff
@@ -46,65 +45,71 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // inflate MapView from layout
```
shellygill: Typo? We're getting the MapView from the layout that's already inflated?

add-graphics-renderer/src/main/java/com/esri/arcgisruntime/sample/addgraphicsrenderer/MainActivity.java
```diff
@@ -46,65 +45,71 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        polygonGeometry.addPoint(20e5, -20e5);
+        polygonGeometry.addPoint(-20e5, -20e5);
+        // solid yellow polygon symbol
+        SimpleFillSymbol polygonSymbol = new SimpleFillSymbol(Color.YELLOW, SimpleFillSymbol.Style.SOLID, null);
+        // create graphic for polygon
+        Graphic polygonGraphic = new Graphic(polygonGeometry.toGeometry());
+        // create graphic overlay for polygon
+        GraphicsOverlay polygonGraphicOverlay = new GraphicsOverlay();
+        // create simple renderer
+        SimpleRenderer polygonRenderer = new SimpleRenderer(polygonSymbol);
```
shellygill: Structure-wise, this sample is all about creating renderer and setting it on a graphics layer, so would be nice if this code was somewhat earlier on, instead of having to scroll all the way to the bottom to find the pertinent lines.

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.1'
+        classpath 'com.android.tools.build:gradle:2.1.0-beta1'
```
shellygill: I know these are Beta samples, but would prefer users only have to deal with one set of Beta issues at a time, so personally I'd think we should use version 2.0.0 here (users always have option to move to beta versions of gradle/android studio themselves).

add-graphics-renderer/src/main/java/com/esri/arcgisruntime/sample/addgraphicsrenderer/MainActivity.java
```diff
@@ -46,65 +45,71 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        for(Point pt : polyPoints){
-            // add graphic to graphics overlay
-            graphics.add(new Graphic(pt));
-        }
-        // create color for graphic
-        int yellow = Color.rgb(255, 255, 0);
+        // point graphic
+        Point pointGeometry = new Point(40e5, 40e5, SpatialReferences.getWebMercator());
+        // yellow diamond point symbol
+        SimpleMarkerSymbol pointSymbol = new SimpleMarkerSymbol(Color.YELLOW, 10, SimpleMarkerSymbol.Style.DIAMOND);
```
shellygill: This color is quite difficult to see on top of the specified basemap - I thought to begin that there wasn't a Point showing up! For the polygon below it's clearer because the polygon takes up a noticeable amount of the map area. Change to Red or similar?

shellygill:
Thanks @doneill - few comments but no specific problems found in the code, very clean!

One comment I had though was that the point of a renderer is to draw multiple symbols using the same Symbol object, instead of having lots of symbols all with a separate instance of Symbol, which might all have the same visual properties. Would it be better to demonstrate this by adding a number of points and then setting a renderer that's used to display all of them, for example? What do you think?

add-graphics-renderer/build.gradle
doneill: Yes, we can do that on the beta-dev branch

add-graphics-renderer/src/main/java/com/esri/arcgisruntime/sample/addgraphicsrenderer/MainActivity.java
doneill: correct, will update the comment

doneill:
Thanks @shellygill I have updated some of your comments and will merge.

Issue #146, opened 2016-04-29
Update display-device-location sample to include runtime permissions


shellygill:
Update Quartz Beta 2 display-device-location sample to include working with Android Marshmallow runtime permissions, using support library.

Request permission in context, when starting the location display (no justification will need to be shown in this case as context of request will be clear to the user).

shellygill:
The PR has been merged with these changes, so closing this issue (a bit tardily!).

PR #147, opened 2016-04-29
Shell0002/location display permissions

shellygill:
@doneill - could you review this update to the quartz-beta branch? This is to address issue #146. The changes are suitable for Beta 2.


doneill:
@shellygill Updates look good, I only have one question before merge, when you run the app for the first time and allow the app to turn on location services then select to stop location service a Toast message opens with the following error:  

location-error

This does not happen if you close the app and restart with the app permissions approved.

shellygill:
@doneill Thanks for review, good catch, this is caused by an issue in our data source for which I've entered an issue; likely something to address outside this sample code.

doneill:
Thanks @shellygill I will merge then.

PR #160, opened 2016-06-01
Updated onStatusChanged listener to use new API for errors

shellygill:
To find out if you've called startAsync on the LocationDisplay but it's failed, you can now check the error object on the event argument passed in (instead of getting this from the data source from the source object from the parameter).

This one-line change updates the sample to use this new API member.

@doneill - could you review this please?


doneill:
Looks good @shellygill merging.

Issue #117, opened 2016-03-18
List items in the OperationalLayers activity disappear


zinfin:
Device:  Nexus 7
Android version: 5.1.1

When long pressing list items in the OperationalLayer activity to change the order in which the layers are added on to the map, one of the two list items disappears. At first, both items act as normal in the re-ordering process, but after a series of rearranging items, one of the items disappears.  First screenshot shows both layers, second screenshot shows screen after moving the 'Census' layer above the 'World Elevations' layer.

<img width="983" alt="screen shot 2016-03-18 at 4 03 45 pm" src="https://cloud.githubusercontent.com/assets/7891400/13894271/42639b58-ed23-11e5-9cfb-195dbc109ed4.png">

![screen shot 2016-03-18 at 3 22 42 pm](https://cloud.githubusercontent.com/assets/7891400/13894239/05179948-ed23-11e5-8822-397bf7aab9f3.png)

doneill:
@zinfin Please verify fix works for you.

zinfin:
@doneill Verified that it's working.

doneill:
Thanks @zinfin closing as done.

PR #168, opened 2016-06-16
refactored to return boolean depending on android version

manishk3189:
@doneill I have created a fix for Issue #117 to address disappearance of list item when re-arranged. Can you please review the same and let me know of changes if required? 

Thanks


doneill:
@manishk3189 Fix looks good, will merge.

PR #171, opened 2016-06-18
Mani8177/route task sample

manishk3189:
Hi @doneill, I have create a sample Find route to address [#1145](https://devtopia.esri.com/runtime/java-android-sdk/issues/1145). Can you please review the same and let me know of changes if required?

Thanks


doneill:
@manishk3189 Please resolve conflicts with this branch before I start to review.

manishk3189:
@doneill The conflicts are resolved now.

find-route/src/main/res/values-w820dp/dimens.xml
```diff
@@ -0,0 +1,7 @@
+<resources>
+    <!-- Example customization of dimensions originally defined in res/values/dimens.xml
+         (such as screen margins) for screens with more than 820dp of available width. This
+         would include 7" and 10" devices in landscape (~960dp and ~1280dp respectively). -->
+    <dimen name="activity_horizontal_margin">64dp</dimen>
```
doneill: This resource appears to be unused.

find-route/src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,6 @@
+<resources>
+    <!-- Default screen margins, per the Android Design guidelines. -->
+    <dimen name="activity_horizontal_margin">16dp</dimen>
+    <dimen name="activity_vertical_margin">16dp</dimen>
```
doneill: Both resources above appear to be unused.

find-route/src/main/java/com/esri/arcgisruntime/sample/findroute/MainActivity.java
```diff
@@ -0,0 +1,292 @@
...
+import com.esri.arcgisruntime.tasks.route.RouteTask;
+import com.esri.arcgisruntime.tasks.route.Stop;
+
+import java.util.List;
+
+public class MainActivity extends AppCompatActivity {
+
+    private static final String TAG = "FindRouteSample";
+    ArcGISVectorTiledLayer mVectorTiledLayer;
+    FloatingActionButton mDirectionFab;
```
doneill: Both field above should be marked as private.

find-route/src/main/java/com/esri/arcgisruntime/sample/findroute/MainActivity.java
```diff
@@ -0,0 +1,292 @@
...
+
+
+        mDirectionFab.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+
+                mProgressDialog.show();
+
+                getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+                getSupportActionBar().setHomeButtonEnabled(true);
```
doneill: `getSupportActionBar)` can be null, you should check for null before calling methods.

doneill:
@manishk3189 I provided some comments for your review.  From a UX perspective, I feel like the directions should interact with the map, e.g. tap on a direction step closes the drawer and highlights/zooms to that step on the map.  I don't think iOS or QT implemented that so I will follow up as I believe we should.

manishk3189:
@doneill I have refactored the sample with the review comments and added zoom/highlight when clicked on direction maneuver from the list. I have also added `setViewpointWithDurationAsync` for better UX experience.

find-route/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,22 @@
...
+
+    <uses-permission android:name="android.permission.INTERNET" />
+
+    <application
+        android:allowBackup="true"
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:supportsRtl="true"
+        android:theme="@style/AppTheme">
+        <activity android:name="com.esri.arcgisruntime.sample.findroute.MainActivity">
```
doneill: Add the following `android:configChanges="orientation|screenSize|uiMode">` to ensure the activity does not restart when screen changes.

doneill:
@manishk3189 The sample is much improved, nice work.  I only have a couple comments left for you, the one above regarding screen orientation and the icons.  

The start and end icons are not pointing to the start/end positions.  You can see this when zoomed in from the directions list:  

Here is the start position: 
start

And the end position, which is a bit better but not exact. 
end

Can you try and get these better positioned?

manishk3189:
Hi @doneill I have pushed changes to address configuration changes and offsets to `PictureMarkerSymbols`. 
The screen privew is much better now.. 
screen shot 2016-07-15 at 1 43 10 pm

--
screen shot 2016-07-15 at 1 43 30 pm

doneill:
@manishk3189 Looks good, will merge.

PR #175, opened 2016-06-20
refactored condition to update feature and apply edits to server

manishk3189:
Hi @doneill, The PR has a fix for issue [#1150](https://devtopia.esri.com/runtime/java-android-sdk/issues/1150). Can you please review the same and let me know of changes if required?

Thanks


doneill:
@manishk3189 Fix looks good, but like #176 we need to update the view point such that there are features in view when app starts.  Please update PR with fix.

manishk3189:
@doneill The initial view point is set to North America now.

PR #247, opened 2016-10-27
Update basemaps sample

doneill:


doneill:
@shellygill I updated the sample to include some vector basemaps, please review.

shellygill:
Looks good @doneill - no changes to suggest. Like the way you can compare vector and raster basemaps of same style.

PR #261, opened 2016-11-17
default authentication sample

doneill:
sample showing the usage of `DefaultAuthenticationChallengeHandler`


doneill:
@willcrick I created a quick example of using `DefaultAuthenticationChallengeHandler` to login into a Portal.  

cc @eric906

doneill:
@XuemingRocks Please review.

authentication-profile/README.md
```diff
@@ -0,0 +1,41 @@
...
+![Authentication Profile App](auth-profile.png) 
+
+## Features
+* AuthenticationManager
+* DefaultAuthenticationChallengeHandler
+* PortalInfo
+* PortalUser
+
+## Developer Pattern
+The `DefaultAuthenticationChallengeHandler` is the default implementation of the `AuthenticationChallengeHandler` interface to handle authentication with a portal.  This is set on the `AuthenticationManager`. To authenticate the user you create a `Portal` object, provide the essential credentials and then asynchronously load the portal as shown below:
```
XuemingRocks: > handle authentication with a portal.

Besides portal `DefaultAuthenticationChallengeHandler` also supports other resources such as layers and tasks. I am not sure if we should explain a bit more here.

authentication-profile/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```java
+// Set the DefaultAuthenticationChallegeHandler to allow authentication with the portal.
+DefaultAuthenticationChallengeHandler handler = new DefaultAuthenticationChallengeHandler(this);
+AuthenticationManager.setAuthenticationChallengeHandler(handler);
+final Portal portal = new Portal("http://www.arcgis.com", true);
```
XuemingRocks: I think we also explain the flag `loginRequired` a bit here or in the source code. When `loginRequired` set to true, the default handler will always prompt for credential while it is false the default handler will only prompt for credential when loading portals that do not allow anonymous access.

authentication-profile/README.md
doneill: I thought about this but since it did not have to do with the sample I left out.  I added, let me know if it reads well.

authentication-profile/src/main/java/com/esri/arcgisruntime/sample/authenticationprofile/MainActivity.java
```diff
@@ -0,0 +1,131 @@
...
+            public void run() {
+                if (portal.getLoadStatus() == LoadStatus.LOADED) {
+                    // Get the portal information
+                    PortalInfo portalInformation = portal.getPortalInfo();
+                    String portalName = portalInformation.getPortalName();
+                    urlText = (TextView)findViewById(R.id.portal);
+                    urlText.setText(portalName);
+
+                    // Get the authenticated portal user
+                    PortalUser user = portal.getUser();
```
XuemingRocks: Should we check if `user` is not null? if our users modify this sample and change `loginRequired` to false, the portal may be loaded without login. Then `user` will be null. Therefore NPE in the following. Maybe it is ok to leave it as it is. your call.

XuemingRocks:
@doneill thanks for the sample. Looks good except a few minor comments. Back to you.

doneill:
Thanks @XuemingRocks I updated based on your comments, let me know what you think.

XuemingRocks:
@doneill looks good 👍

PR #317, opened 2017-05-04
Trev8939/blend renderer

TADraeseke:
Next release of the SDK will no longer allow creating a new `RasterLayer`s from a `Raster` which is already used to create an existing `RasterLayer`.  Fix creates new `RasterLayer`s when updating the renderer.

### DIFF
--- a/blend-renderer/src/main/java/com/esri/arcgisruntime/sample/blendrenderer/MainActivity.java
+++ b/blend-renderer/src/main/java/com/esri/arcgisruntime/sample/blendrenderer/MainActivity.java
@@ -133,8 +133,8 @@ public class MainActivity extends AppCompatActivity implements ParametersDialogF
   private void updateRenderer() {
     ColorRamp colorRamp = mColorRampType != ColorRamp.PresetType.NONE ? new ColorRamp(mColorRampType, 800) : null;
     // if color ramp is not NONE, color the hillshade elevation raster instead of using satellite imagery raster color
-    RasterLayer mRasterLayer = colorRamp != null ? new RasterLayer(mElevationRaster) : new RasterLayer(mImageryRaster);
-    mMapView.getMap().setBasemap(new Basemap(mRasterLayer));
+    RasterLayer rasterLayer = colorRamp != null ? new RasterLayer(mElevationRaster) : new RasterLayer(mImageryRaster);
+    mMapView.getMap().setBasemap(new Basemap(rasterLayer));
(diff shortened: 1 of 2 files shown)


blend-renderer/src/main/java/com/esri/arcgisruntime/sample/blendrenderer/MainActivity.java
```diff
@@ -29,6 +29,10 @@
 
   private MapView mMapView;
 
+  private File mImageFile;
+
+  private File mElevationFile;
+
   private Raster mImageryRaster;
 
   private Raster mElevationRaster;
```
doneill: The `Raster` fields above should be converted to local variables.

doneill: Reduce the amount of spaces between your globals.

TADraeseke:
@doneill global -> local variables done.

As to the whitespace, Shelly's said I ought to start using the [auto-format file](https://devtopia.esri.com/runtime/java-common-api/wiki/Java-and-Android-code-guidelines) (which introduces that extra whitespace between global variables). If there's a different format on samples, happy to change it!

doneill:
@TADraeseke The guideline doc you reference needs to be updated, although I don't see where it states every variable needs a line space between them.  Otherwise updates look good to merge.

TADraeseke:
@doneill When Code > Reformat Code is run using the ESRI_RTC_Java scheme, it adds all those line spaces. I can remove them manually or merge as is, depending on your preference.

doneill:
@TADraeseke Hmm, ok, yeah we will need to update that as it may be ok for SDK code we can pass more globals around in app facing code.  Block them up for now and I will look to update our public facing format. 

```java
  private MapView mMapView;
  private Raster mImageryRaster;
  private Raster mElevationRaster;
  private SlopeType mSlopeType;

  private int mAltitude;
  private int mAzimuth;
  private double mZFactor;
  
  private ColorRamp.PresetType mColorRampType;
  private double mPixelSizeFactor;
  private double mPixelSizePower;
  private int mOutputBitDepth;
  private FragmentManager mFragmentManager;
```

TADraeseke:
@doneill Reformatted, including blocking globals. Reformatting code also moved listener in ParametersDialogFragment to the bottom of the code.

doneill:
@TADraeseke good to merge.

PR #350, opened 2017-06-17
update related features sample

doneill:


doneill:
@rchintapalli Ready for your review.

doneill:
@rchintapalli No, you should be able to update the attribute multiple times from a single selection. I am able to do the following:  

1. Tap on a feature
2. Get Callout with Related Feature Annual Visitors selected in drop down. 
3. Select a different value from drop down
4. Sync to server
5. Callout updated with new value, select a different value from drop down
6. Sync to server
7. Verify values from service as updated with second update.  

Where do you have issues in the above  workflow?

rchintapalli:
@doneill 
When I  Select value 10000-50000 it updates the service, and then select 0-1000, no update request sent now. Looks like with other values it is working fine and only when we try to go back to 0-1000, I see the issue

doneill:
@rchintapalli I found the inconsistency and provided a fix, let me know if it works for you now.

rchintapalli:
@doneill Looks good. works fine now

PR #370, opened 2017-08-15
Trev8939/feature collection query

TADraeseke:
A sample which demonstrates use of feature collections and querying.

https://devtopia.esri.com/runtime/java-android-sdk/issues/1376


TADraeseke:
@doneill If you have time, this should be a fairly quick/straight forward review

PR #375, opened 2017-08-30
Trev8939/generateGeodatabase

TADraeseke:
@doneill available to review this?

Issues which would require your attention:

- Provide link for SanFrancisco.tpk from AGOL (as per provision your device in README)

- Design decision of what to do with the local Geodatabase file (should it persist on closing?, be deleted? be stored in temp cache?)

- Should the sample allow for multiple downloads of geodatabase, like in the [TileCache sample](https://github.com/Esri/arcgis-runtime-samples-android/tree/master/export-tiles) or is once enough? At the moment the app removes the generate button once a local geodatabase has loaded.


doneill:
@TADraeseke I notice the data has not been add to AGOL, I will do that tomorrow and review.

generate-geodatabase/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="colorPrimary">#2196F3</color>
+    <color name="colorPrimaryDark">#1976D2</color>
+    <color name="colorAccent">#4CAF50</color>
+    <color name="dimBackground">#BF000000</color>
```
doneill: `dimBackground` resource appears to be unused.

generate-geodatabase/src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,5 @@
+<resources>
+    <!-- Default screen margins, per the Android Design guidelines. -->
+    <dimen name="activity_horizontal_margin">16dp</dimen>
+    <dimen name="activity_vertical_margin">16dp</dimen>
```
doneill: both resources appear to be unused.

generate-geodatabase/src/main/res/values-w820dp/dimens.xml
```diff
@@ -0,0 +1,6 @@
+<resources>
+    <!-- Example customization of dimensions originally defined in res/values/dimens.xml
+         (such as screen margins) for screens with more than 820dp of available width. This
+         would include 7" and 10" devices in landscape (~960dp and ~1280dp respectively). -->
+    <dimen name="activity_horizontal_margin">64dp</dimen>
```
doneill: resource appears to be unused

generate-geodatabase/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,13 @@
+<resources>
+    <string name="app_name">Generate Geodatabase</string>
+    <string name="world_street_map">https://sampleserver6.arcgisonline.com/arcgis/rest/services/World_Street_Map/MapServer</string>
```
doneill: `world_street_map` string resource is unused.

generate-geodatabase/src/main/res/drawable/preview_mask.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
```
doneill: This resource appears unused.

doneill:
@TADraeseke I updated the README with data links, please verify it works for you.

generate-geodatabase/src/main/java/com/esri/arcgisruntime/sample/generategeodatabase/MainActivity.java
```diff
@@ -0,0 +1,247 @@
...
+  @Override
+  protected void onResume() {
+    super.onResume();
+    mMapView.resume();
+  }
+
+  @Override protected void onDestroy() {
+    super.onDestroy();
+    if (isFinishing()) {
+      // TODO delete local database???
```
doneill: Remove TODO comment

TADraeseke:
@doneill Thanks for review. All PR fixes taken care of. I just realized I've been using Analyze -> Inspect Code on .java file instead of the module, which explains why I kept missing the unused resources. I should catch more of these going forward!


Just wanted to bring up one more time (in case it gets forgotten about). Is there a feeling on storing local info on the Android device when generated through a `job` like this? Android best practice seems to suggest we should be using the [CacheDir](https://developer.android.com/reference/android/content/Context.html#getCacheDir()). Alot to be said for letting Android OS take care of files. However, it maybe makes samples less clear, in that the developer can't as easily navigate to a local folder to see that data has been stored locally?

I only ask since this sample is directly affected.

[This code](https://github.com/Esri/arcgis-runtime-samples-android/pull/375/files#diff-2623bebe0eb0054e2ccec8da7dcd1322R235) is pretty ugly, as onDestroy is not guaranteed to be called within Android lifecycles:
```
@Override protected void onDestroy() {
    super.onDestroy();
    if (isFinishing()) {
      boolean deleted = new File(mLocalGeodatabasePath).delete();
      if (deleted) {
        Log.i(TAG, "Local Geodatabase deleted.");
      } else {
        Log.e(TAG, "Failed to delete local Geodatabase.");
      }
    }
  }
```

doneill:
@TADraeseke We persist/provision data on internal storage when apps are intended to be used offline. The context of this sample is that while online we can generate a geodatabase locally.  A use case for that context would be to persist the data but since we don't take it that far I agree that we should take advantage of caching data.  Go ahead and refactor to support use of `getCacheDir()` and save the `File` as a temporary cache file.  We should also ensure we limit size of the geodatabase to within a reasonable limit, e.g. < 5MB.

TADraeseke:
@doneill Have refactored to use the CacheDir, but don't know the best way to limit geodatabase size before its generated by the job. Any tips in that regard?

Maximum for this sample (taking the whole of san francisco) is 1.3 mb

PR #412, opened 2017-11-09
Trev8939/rend mode scene camera fix

TADraeseke:
Change roll on zoomedIn camera constructor from 90 to 0.


TADraeseke:
Thanks @shellygill!

PR #416, opened 2017-11-20
Trev8939/symbolize shapefile

TADraeseke:
Add a sample for changing symbolization of a shapefile


TADraeseke:
Thanks @shellygill. Fixed as per your comments!

PR #420, opened 2017-11-23
Trev8939/wms layer

TADraeseke:
Adds wms layer url sample


java/wms-layer-url/src/main/java/com/esri/arcgisruntime/sample/wmslayerurl/MainActivity.java
```diff
@@ -0,0 +1,68 @@
...
+        // inflate MapView from layout
+        mMapView = findViewById(R.id.mapView);
+        // create a map with the BasemapType topographic
+        ArcGISMap map = new ArcGISMap(Basemap.Type.IMAGERY, 2.0, 18.0, 3);
+        // set the map to be displayed in this view
+        mMapView.setMap(map);
+
+        // hold a list of uniquely-identifying WMS layer names to display
+        List<String> wmsLayerNames = new ArrayList<>();
+        wmsLayerNames.add("0");
```
shellygill: This looks like an ID rather than a layer 'name' - is this the right value?

java/wms-layer-url/src/main/java/com/esri/arcgisruntime/sample/wmslayerurl/MainActivity.java
```diff
@@ -0,0 +1,68 @@
...
+
+        // create a new WMS layer displaying the specified layers from the service
+        WmsLayer wmsLayer = new WmsLayer(getString(R.string.wms_layer_url), wmsLayerNames);
```
shellygill: This service does not seem to be working currently - cant access in a browser, and loading layer fails with "invalid XML" error.

java/wms-layer-url/src/main/java/com/esri/arcgisruntime/sample/wmslayerurl/MainActivity.java
TADraeseke: I've just replicated this from initial .NET implementation, but I can clarify in the next samples scrum

TADraeseke: Ok, so apparently that is the name of the layer. The WMS service 2 layers with the inspiring names '0' and '1' :P

https://certmapper.cr.usgs.gov/arcgis/services/geology/africa/MapServer/WMSServer?request=GetCapabilities&service=WMS

java/wms-layer-url/src/main/java/com/esri/arcgisruntime/sample/wmslayerurl/MainActivity.java
TADraeseke: It's an external service hosted by the USGS, which is intentional, which means it may occasionally go down. This is meant to show support for non ESRI managed WMSs.

This does mean if USGS changes the url or closes the service the sample won't work though...

shellygill:
Thanks @TADraeseke - it's working OK for me now. Given your comments about intent of sample and external service above, worth adding a note to the readme to this effect before merging I think. Apart from that looks good to merge.

PR #422, opened 2017-11-27
Trev8939/location los

TADraeseke:
Adds sample demonstrating the use of LocationLineOfSight


java/location-line-of-sight/src/main/java/com/esri/arcgisruntime/sample/locationlineofsight/MainActivity.java
```diff
@@ -0,0 +1,115 @@
...
+        // get a screen point from the motion event
+        android.graphics.Point screenPoint = new android.graphics.Point(Math.round(motionEvent.getX()),
+            Math.round(motionEvent.getY()));
+
+        // convert the screen point to a scene point
+        Point scenePoint = mSceneView.screenToBaseSurface(screenPoint);
+
+        if (mObserverLocation == null) {
+          // set the observer location
+          mObserverLocation = new Point(scenePoint.getX(), scenePoint.getY(), scenePoint.getZ() + mZOffset, scenePoint.getSpatialReference());
```
shellygill: You can get an uncaught error here (and similar line below) as scenePoint can be null - e.g. if you tap off the horizon of the earth (e.g. there's no valid location to return).

TADraeseke: Good catch! Thanks @shellygill

TADraeseke:
@shellygill fix added to catch for point selection not on the surface

shellygill:
Thanks @TADraeseke changes look good.

PR #454, opened 2018-02-26
Trev8939/preplanned map

TADraeseke:
Sample demonstrating use of the download preplanned map workflow.


tschie:
@TADraeseke Did you say you might refactor this one? I will hold off re-reviewing for now.

PR #496, opened 2018-06-13
Change elevation service

eri9600:
@TADraeseke can you review this when you can.
I am changing the elevation service as stated in the slack channel. 
Thanks!

### DIFF
--- a/java/distance-measurement-analysis/src/main/res/values/strings.xml
+++ b/java/distance-measurement-analysis/src/main/res/values/strings.xml
@@ -1,7 +1,6 @@
 <resources>
     <string name="app_name">Distance Measurement Analysis</string>
-    <string name="world_terrain_3d">http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer</string>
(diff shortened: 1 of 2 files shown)


java/distance-measurement-analysis/src/main/res/values/strings.xml
```diff
@@ -1,7 +1,6 @@
 <resources>
     <string name="app_name">Distance Measurement Analysis</string>
-    <string name="world_terrain_3d">http://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer</string>
-    <string name="MNT_IDF">https://tiles.arcgis.com/tiles/d3voDfTFbHOCRwVR/arcgis/rest/services/MNT_IDF/ImageServer</string>
+    <string name="world_terrain_3d">"http://scene.arcgis.com/arcgis/rest/services/BREST_DTM_1M/ImageServer"</string>
```
TADraeseke: Good service to use as its used in several other scene samples, however its not a world terrain surface (so maybe just change the name to "elevation_service" or "brest_dtm" or something. Also, doesn't need the quotes around it.

PR #523, opened 2018-08-10
change button to switch and update screenshot

TADraeseke:
Update sample to use a switch instead of a button to toggle between extrusion expressions. Could you review when you have time please @eri9600 ? Thank you!


eri9600:
@TADraeseke Good idea to put a switch instead. 
The only thing i can see is that the copyRight could be updated to 2018. Otherwise, everything looks good!

TADraeseke:
Thanks @eri9600 - I've merged as is. I think we usually just leave the copyright to the year the sample was made :)

PR #530, opened 2018-09-26
fixes for Update Related Features

TADraeseke:
Hey Hardik! Had a look at this sample as per: https://devtopia.esri.com/runtime/common-samples/issues/861

I can see nothing in the code that handles clicks on the non-green layer. So the problem we're seeing is by design. I've cleaned up the code, fixed a deprecation issue and updates the README could you please review/verify?

Thanks!

### DIFF
--- a/java/update-related-features/build.gradle
+++ b/java/update-related-features/build.gradle
@@ -11,11 +11,17 @@ android {
         versionCode rootProject.ext.versionCode
         versionName rootProject.ext.versionName
     }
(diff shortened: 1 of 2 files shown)


hardikc07:
@TADraeseke what should be the expected behavior ? as i am still seeing similar behavior ( i mean click on yellow color and alert popups for a sec and disappear ) , i have tried running sample with above branch

PR #542, opened 2018-10-17
issue934 red pin and laderposition need to be set correct

priy1289:


java/offline-geocode/src/main/java/com/esri/arcgisruntime/sample/offlinegeocode/MainActivity.java
```diff
@@ -224,7 +224,6 @@ public void run() {
     mPinSourceSymbol.setWidth(20);
     mPinSourceSymbol.loadAsync();
     mPinSourceSymbol.setLeaderOffsetY(45);
-    mPinSourceSymbol.setOffsetY(-48);
```
TADraeseke: I think what @hardikc07 is looking for here is .setOffsetY(-45). Keeping the callout at the top of the pin allows the user to drag the pin around to reverse geocode.

TADraeseke: This was not part of the issue, but I think we should probably fix it anyways...

This code is showing a deprecated way of creating a picture marker symbol (Line 222). Could you change it to look something like: https://github.com/Esri/arcgis-runtime-samples-android/blob/65533272642bd366e31af709c250063ded30219a/java/find-route/src/main/java/com/esri/arcgisruntime/sample/findroute/MainActivity.java#L235

Also the line    `  //[DocRef: Name=Picture Marker Symbol Drawable-android, Category=Fundamentals, Topic=Symbols and Renderers]` is a copy paste error and can be removed. FYI - its a tag for a script we run which pulls out code snippets used in our documentation.

priy1289: I spoke to Hardik and we see 2 problems here:
1. the red pin location and touch are not in sync. the red pin is always some points lower on Y axis than the touch. This is set by the offsetY on mPinSourceSymbol.
2. I agree to your suggestion that the callout should be placed bit higher than the red pin. 
This is set by the onLongPress function, in the variable screenpoint. Line 459. 

To address the above two points, I set the mPinSourceSymbol.setOffsetY to -25 and screenPoint Y coordinate to -75 [e.getY()-75]. Is there a better way to do this? since it can vary for different phone screen sizes?

This way i am able to see the red pin on the touch point and also the callout on top of the red pin.

priy1289:
Closing this pull request, as this was addressed as part of https://github.com/Esri/arcgis-runtime-samples-android/pull/545

PR #545, opened 2018-10-22
Trev8939/clean offline geocode

TADraeseke:
Hi @priy1289 - was easier just to create this PR against master. Can you review and let me know what you think. Specifically:

- Does this rework of the sample clarify the parts of the API we're trying to show (geocode/reverse geocode from offline data)
- Does it seem like reasonable java/android code

Thanks!


java/offline-geocode/src/main/java/com/esri/arcgisruntime/sample/offlinegeocode/MainActivity.java
```diff
@@ -363,241 +325,90 @@ protected void onResume() {
...
 
-    float dX, dY;
-
-    public DragTouchListener(Context context, MapView mapView) {
+    DragTouchListener(Context context, MapView mapView) {
       super(context, mapView);
     }
 
     @Override
     public boolean onTouch(View view, MotionEvent event) {
```
priy1289: Looks like onTouch function is not needed. It is not called from anywhere in the code.

java/offline-geocode/src/main/java/com/esri/arcgisruntime/sample/offlinegeocode/MainActivity.java
```diff
@@ -363,241 +325,90 @@ protected void onResume() {
...
   protected void onDestroy() {
-    super.onDestroy();
     mMapView.dispose();
+    super.onDestroy();
   }
 
+  /**
+   * Define a listener to handle drag events.
+   */
   private class DragTouchListener extends DefaultMapViewOnTouchListener {
```
priy1289: Not sure if we are really using the DragTouchListener class. I believe this is to drag the pin to desired location? Its not working in this sample. hence the onSingleTapConfirmed function really doesn't do anything. Please correct me if I misunderstood here.

priy1289: ok got it, it works on single tap the pin and then tap hold and drag the pin.

PR #552, opened 2018-11-08
Priya1289/issue956

priy1289:
fix for app crash


priy1289:
the app crashed because it couldn't handle very large data. It crashes for any group fields selected that returns a very large dataset. Intent putextra and getextra couldn't handle this load, and app crashed. Fixed using another suggested way to pass data between different activities -with a public static field.

priy1289:
Thanks @TADraeseke !

PR #559, opened 2018-11-21
Priy1289/dictionary renderer graphics overlay

priy1289:
sample for dictionary renderer with graphics overlay.


priy1289:
opened another one.

PR #581, opened 2019-02-13
Luk10114/1193 fix samples on rotation/analyze hotspots

LDuncAndroid:
See https://devtopia.esri.com/runtime/common-samples/issues/1193 for context


java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/AnalyzeHotspotsFragment.java
```diff
@@ -0,0 +1,276 @@
...
+import com.esri.arcgisruntime.loadable.LoadStatus;
+import com.esri.arcgisruntime.mapping.ArcGISMap;
+import com.esri.arcgisruntime.mapping.Basemap;
+import com.esri.arcgisruntime.mapping.Viewpoint;
+import com.esri.arcgisruntime.mapping.view.MapView;
+import com.esri.arcgisruntime.tasks.geoprocessing.*;
+
+import java.util.concurrent.ExecutionException;
+
+public class AnalyzeHotspotsFragment extends Fragment {
```
gunt0001: I think we should still keep the actual geoprocessing logic in the MainActivity. At least in the context of the samples I think it makes sense to keep it in MainActivity. That way this code would also be immediately visible to the user in the SampleViewer. @TADraeseke would be able to give you some more details on SampleViewer requirements.

I think it's ok to factor out the dialogs into their own DialogFragment implementations.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/DateRangeDialogFragment.java
```diff
@@ -0,0 +1,214 @@
...
+
+  private Date mMinDate;
+
+  private Date mMaxDate;
+
+  private OnAnalyzeButtonClickListener mOnAnalyzeButtonClickListener;
+
+  private String title;
+
+  private String buttonText;
```
gunt0001: Could we fix the inconsistent naming of fields as part of these changes as well? I think for samples we use `mXYZ` style. @TADraeseke please confirm.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,316 +16,41 @@
...
-   * ArcGISMapImageLayer to the map and resets the Viewpoint of the MapView.
-   *
-   * @param from string which holds a date
-   * @param to   string which holds a date
-   */
-  private void analyzeHotspots(final String from, final String to) {
-    // cancel previous job request
-    if (mGeoprocessingJob != null) {
-      mGeoprocessingJob.cancel();
+    if (savedInstanceState == null) {
```
gunt0001: Could there be a case where `savedInstanceState` would not be null? If yes, how would we handle that case?

LDuncAndroid: We don't want to re-add the `Fragment` as it's already been retained and restored by the OS. If the `savedInstanceState` isn't null, we're already showing the `Fragment` instance.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/AnalyzeHotspotsFragment.java
gunt0001: So we have two choices of retaining state between configuration changes:
1. Implement a Fragment and call `setRetainInstance(true)` on it.
2. Keep the logic in the MainActivity and specify `android:configChanges="orientation| screenSize |keyboardHidden"` for it in the manifest.

Since our samples usually don't have complicated UI that would have to be updated when we handle configuration changes ourselves, I would go for option 2 as this would keep the logic in the MainActivity, which means that code would also show up in the Sample Viewer. @TADraeseke - can you please comment on whether that's a requirement for the Sample Viewer - i.e. that the main code of the sample needs to be in MainActivity in order to show up in the Sample Viewer?

From my POV the Fragment would make sense if we had to show different UI components in the MainActivity but usually our samples aren't that complex.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
gunt0001: ok

TADraeseke:
Just had a discussion with @LDuncAndroid - I think the progress dialog fragment and date picker fragments are an obvious win here. They add some code, but critically they add them to classes away from runtime API code (so mean less UI code intermixed with runtime API code).

However, long running operations like a job are not unique to the Analyze Hotspots sample and the cost in terms of extra code is higher and as @gunt0001 said it means devs are digging through different classes to find runtime API code (which should be front and center).

This probably highlights the need for us to have a lifecycle aware/device config change/etc sample we can point to for those best practices. We just don't want to spend our time and lines of code replicating those patterns in every sample which includes a long running operation.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/AnalyzeHotspotsFragment.java
TADraeseke: Sorry its taken me so long to get to this. Thanks both for this. This is a long running issue with samples...where to find the balance between a primary goal of readability of runtime code and a secondary goal of showing best practice with runtime API within Android's API.

I also like option 2, and we used to use this on most samples as a convenient get out of jail free card...but at some point we thought this was showing bad practice and removed it from all of our samples.

On one hand we have simple API calls for handling long running jobs (pause, cancel and progress listener etc) which we want to show being used. However, if they're not handled in a lifecycle aware way, they just run off into an eventual error (if the device rotates mid job for example).

I personally think we should use a cheat like configChanges in this case, and implement one sample (maybe even this one) as a demo of a life cycle best practice with long running parts of our APIs (jobs in particular).

gunt0001: 👍 for option 2 and one (or more) separate sample(s) that specifically discuss any aspects of life cycle events. These new samples could incorporate toolkit components that we might implement, for better support of restoring MapView/SceneView state after a destroy/re-create (MapFragment/MapViewModel).

LDuncAndroid:
@TADraeseke updated this to reflect the changes we've discussed.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,39 +16,242 @@
...
-      return fragment;
-    } else {
-      return AnalyzeHotspotsFragment.newInstance();
+  @Override
+  public void onResume() {
+    super.onResume();
+    mMapView.resume();
+
+    // If the layer has already been created, add it to the map view
+    if (mHotspotMapImageLayer != null) {
```
TADraeseke: do we still need these calls if we're not doing the geoprocessing in a fragment and/or handling config changes in manifest?

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,39 +16,242 @@
...
+        // set the map viewpoint to the MapImageLayer, once loaded
+        mMapView.setViewpoint(new Viewpoint(mHotspotMapImageLayer.getFullExtent()));
+      }
     }
   }
+
+  @Override
+  public void onPause() {
+    // We have to disassociate the layer from the map before reattaching it when the Fragment is recreating
+    mMapView.getMap().getOperationalLayers().remove(mHotspotMapImageLayer);
```
TADraeseke: same as above

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,39 +16,242 @@
...
+import com.esri.arcgisruntime.tasks.geoprocessing.GeoprocessingResult;
+import com.esri.arcgisruntime.tasks.geoprocessing.GeoprocessingString;
+import com.esri.arcgisruntime.tasks.geoprocessing.GeoprocessingTask;
 
 public class MainActivity extends AppCompatActivity implements DateRangeDialogFragment.OnAnalyzeButtonClickListener,
     ProgressDialogFragment.OnProgressDialogCancelListener {
 
+  private static final String TAG = MainActivity.class.getSimpleName();
+
+  // String used to query input for GeoprocessingString
```
TADraeseke: make all comments lower case

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,39 +16,242 @@
...
+            GeoprocessingResult geoprocessingResult = mGeoprocessingJob.getResult();
+            mHotspotMapImageLayer = geoprocessingResult.getMapImageLayer();
+
+            addLayerToMapView(mHotspotMapImageLayer, mMapView);
+
+          } else if (mCancelled) {
+            Toast.makeText(MainActivity.this, "Job canceled.", Toast.LENGTH_SHORT).show();
+            Log.i(TAG, "Job cancelled.");
+          } else {
+            Log.e(TAG, "Job did not succeed!");
```
TADraeseke: can we get and report the error from the job itself too

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,39 +16,242 @@
...
+        e.printStackTrace();
+      }
+    });
+  }
+
+  private void addLayerToMapView(final Layer layer, final MapView mapView) {
+    // add the new layer to the map
+    mapView.getMap().getOperationalLayers().add(layer);
+
+    // Add a listener to set the viewpoint of the map view when it is finished loading only if the layer has not already
```
TADraeseke: lowercase A

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
```diff
@@ -16,39 +16,242 @@
...
+            findProgressDialogFragment().dismiss();
+          }
+        });
+      } catch (InterruptedException | ExecutionException e) {
+        e.printStackTrace();
+      }
+    });
+  }
+
+  private void addLayerToMapView(final Layer layer, final MapView mapView) {
```
TADraeseke: maybe a short method comment to sum up rather than inline comments below?

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
LDuncAndroid: The idea behind adding the config changes was to allow us to persist these over rotation, so these fields are still being persistsed.

LDuncAndroid: * we check if it's null to check if it's been created or not.

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
LDuncAndroid: https://github.com/Esri/arcgis-runtime-samples-android/pull/581#discussion_r267281402

java/analyze-hotspots/src/main/java/com/esri/arcgisruntime/sample/analyzehotspots/MainActivity.java
TADraeseke: But aren't we only ever hitting onResume once now, when the app launches?

LDuncAndroid: Right enough, I was under the impression the lifecycle was still happening. Will remove.

PR #586, opened 2019-02-18
Trev8939/offline map local basemap

TADraeseke:
@LDuncAndroid - Can you take a look at this please? Just a first pass to see if I've accurately replicated the dialog fragment patterns. Thanks!


java/generate-offline-map-with-local-basemap/src/main/java/com/esri/arcgisruntime/sample/generateofflinemapwithlocalbasemap/MainActivity.java
```diff
@@ -0,0 +1,384 @@
...
+  @Override public void onNegativeClick() {
+    // call generate offline map with parameters which contain an empty string for reference basemap directory
+    generateOfflineMap();
+  }
+
+  /**
+   * Find and return the progress dialog fragment.
+   * @return the progress dialog fragment.
+   */
+  private ProgressDialogFragment findProgressDialogFragment() {
```
LDuncAndroid: This is used so we don't display two instances of the same `Fragment`. You could do this for the loading dialog too.

java/generate-offline-map-with-local-basemap/src/main/java/com/esri/arcgisruntime/sample/generateofflinemapwithlocalbasemap/ProgressDialogFragment.java
```diff
@@ -0,0 +1,98 @@
...
+    args.putString(ARGS_MESSAGE, message);
+    args.putString(ARGS_CANCEL, cancel);
+    fragment.setArguments(args);
+    return fragment;
+  }
+
+  @Override public void onCreate(@Nullable Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    // prevent re-creation during configuration chance to allow us to dismiss this DialogFragment
+    setRetainInstance(true);
```
LDuncAndroid: I'm starting to question whether this is a good idea. Generally we shouldn't use it but I understand I left it in my previous PR. Will investigate.

LDuncAndroid: I've done a refactor of the Analyze Hotspots ticket again and move quite alot around to attempt to retain state over configuration changes. Happy to leave this here and we can discuss this further.

java/generate-offline-map-with-local-basemap/src/main/java/com/esri/arcgisruntime/sample/generateofflinemapwithlocalbasemap/MainActivity.java
TADraeseke: Ahh ok. Makes sense. I know this is definitely more robust, but is there a way in this sample that two progress or alert dialogs could reasonably be created with the current UI? If not, I'm tempted to refactor away from the find___DialogFragment() type methods (rather than implement it for the AlertDialog fragment.

java/generate-offline-map-with-local-basemap/src/main/java/com/esri/arcgisruntime/sample/generateofflinemapwithlocalbasemap/ProgressDialogFragment.java
TADraeseke: I actually kind of like the setRetainInstance(true) code, even if it involved a work around of an Android API bug...

java/generate-offline-map-with-local-basemap/src/main/java/com/esri/arcgisruntime/sample/generateofflinemapwithlocalbasemap/MainActivity.java
LDuncAndroid: I think if you spammed the trigger that created and commits the dialog, there's a chance it could happen and I can't remember off the top of my head but I think an exception is thrown if you try to re-add the same fragment with the same tag to the screen.

This method is handy for we want to update the progress UI, as we'll need to keep trying to find by tag.

PR #590, opened 2019-02-25
Luk10114/192 auth with oauth2

LDuncAndroid:
https://devtopia.esri.com/runtime/common-samples/issues/192


LDuncAndroid:
@gunt0001

LDuncAndroid:
@gunt0001 I have added notes to README and comments in source files highlighting that we do not recommend storing Access Tokens in `SharedPreferences`

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,343 @@
...
+    webView.settings.javaScriptEnabled = true
+    webView.visibility = View.VISIBLE
+  }
+
+  /**
+   * Generate the Intent that launches the MainActivity with the new auth code
+   *
+   * @param uri instance of Uri generated from URL that OAuth webpage redirects to after successful login
+   */
+  private fun generateAuthIntent(uri: Uri): Intent {
```
TADraeseke: I guess its this function here where we need to generate a unique intent flag to not accidently call other similarly IDed Activities?

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,343 @@
...
+
+fun AppCompatActivity.logToUser(logMessage: String) {
+  Log.d(this::class.java.simpleName, logMessage)
+  Toast.makeText(this, logMessage, Toast.LENGTH_LONG).show()
+}
+
+/**
+ * SharedPreferences extensions
+ *
+ * The user's access token should be treated with the upmost of security, and as such, we do not recommend storing the user's
```
TADraeseke: utmost

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,343 @@
...
+import com.esri.arcgisruntime.security.UserCredential
+import kotlinx.android.synthetic.main.activity_main.*
+
+/**
+ * This sample demonstrates how to authenticate with ArcGIS Online (or your own portal) using OAuth2 to access secured
+ * resources (such as private web maps or layers). Accessing secured items requires a login on the portal that hosts them
+ * (an ArcGIS Online account, for example). This sample utilizes Android WebView to show theOAuth sign-in page in a dialog.
+ *
+ * The user's access token should be treated with the upmost of security, and as such, we do not recommend storing the user's
+ * Access Token in SharedPreferences as SharedPreferences are not encrypted. We have created this sample to highlight the use of our APIs.
```
TADraeseke: Lets refer to it as ArcGISRuntime API, rather than our API (if thats the one you mean).

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,343 @@
...
+fun AppCompatActivity.logToUser(logMessage: String) {
+  Log.d(this::class.java.simpleName, logMessage)
+  Toast.makeText(this, logMessage, Toast.LENGTH_LONG).show()
+}
+
+/**
+ * SharedPreferences extensions
+ *
+ * The user's access token should be treated with the upmost of security, and as such, we do not recommend storing the user's
+ * Access Token in SharedPreferences as SharedPreferences are not encrypted. We have created this sample to highlight the use of our APIs.
```
TADraeseke: as above, don't use "our API"

LDuncAndroid:
> In addition to sorting out the intent flag/ID

@TADraeseke looks like the ArcGIS Online Portal won't let us use a host/scheme combo that doesn't match `my-ags-app://auth`, each time I've tried to change it I get this error.

image

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: What we define in the manifest must be passed to the Portal as part of the request to Auth.
```
 <intent-filter>
                <action android:name="android.intent.action.VIEW" />

                <category android:name="android.intent.category.DEFAULT" />
                <category android:name="android.intent.category.BROWSABLE" />

                <data
                    android:host="auth"
                    android:scheme="my-ags-app" />
            </intent-filter>
```

`my-ags-app://auth`

So what we do is when we're creating our `OAuthConfiguration`, we pass `my-ags-app://auth` as the re-direct URI so when a user auths successfully, the user is redirected to that URI which our app has declared as it can handle in it's manifest.

If we changed the URI here, which we could do but we'd need to add another `intent-filter` to the manifest, we'd still be redirected to `my-ags-app://auth` as the portal is forcing this:
https://github.com/Esri/arcgis-runtime-samples-android/pull/590#issuecomment-473006032

TADraeseke:
Thanks @LDuncAndroid - Good work sorting the custom URL scheme! LGTM!

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,348 @@
...
+  private fun beginOAuth() {
+    // get the authorization code by sending user to the authorization screen
+    val authorizationUrl = OAuthTokenCredentialRequest.getAuthorizationUrl(
+      oAuthConfig.portalUrl, oAuthConfig.clientId, oAuthConfig.redirectUri, ACCESS_TOKEN_EXPIRY_MINS
+    )
+
+    // create an Intent to attempt to handle the authorization URL
+    with(Intent(Intent.ACTION_VIEW, Uri.parse(authorizationUrl))) {
+      this.resolveActivity(packageManager)?.let {
+        // this identifier ensures that the browser will attempt to reuse the same window each time the application
```
PuneetEsri: I am not clear on what is the intent of this sample. The java version that worked on for ArcGIS indoors team was to illustrate how to use custom challenge handler to use OAuth in a WebView. 

If the use wants to launch the login window in a separate browser, they can just use the `DefaultChallengeHandler`.  Or if they want to preemptively challenge the user they can use `OAuthLoginManager.launchOAuthBrowserPage​()` 
https://developers.arcgis.com/android/latest/api-reference/reference/com/esri/arcgisruntime/security/OAuthLoginManager.html

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,348 @@
...
+   * @return the AuthenticationChallengeResponse indicating which action to take
+   */
+  override fun handleChallenge(authenticationChallenge: AuthenticationChallenge?): AuthenticationChallengeResponse {
+    authenticationChallenge?.let { authChallenge ->
+
+      try {
+        // if SharedPreferences has an access token
+        sharedPreferences.accessToken?.let { accessToken ->
+          // USER_CREDENTIAL_CHALLENGE is issued when attempt to use access token fails
+          if (authChallenge.type == AuthenticationChallenge.Type.USER_CREDENTIAL_CHALLENGE) {
```
PuneetEsri: This should not be happening if the the Portal is secured by OAuth we should not be getting a `USER_CREDENTIAL_CHALLENGE` 
@LDuncAndroid can you share the repro scenario ?

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: Double check with Xeuming but I believe the intent was to show usage a of Custom Challenge Handler but as Xeuming pointed out before, the use of a `WebView` is a security issue. So I showed how to use with a web browser. Also, I believe they wanted it in Kotlin.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: It is issued when the access token is invalid/expired.

LDuncAndroid:
@PuneetEsri , I believe I've addressed your issues. Can you review again? Thanks.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
PuneetEsri: My Point is that this is a Sample, which our users use as a starting point. Most will copy/paste this code. Here we are not using our API that we provide. We have an issue to fix our API with this call to launch the same browser window. We should be making calls to our API here and not showing the user how to launch the browser window in Android. 

Also This logic here to launch the browser window does not make sense, as I don't think there any modern phones out there that are shipped without a web browser. So why don't we just restrict this sample to just show the use of Webview and have a separate Sample that shows the use of DefaultChallengeHandler that dishes out the OAuth prompt in a web browser.

LDuncAndroid:
@TADraeseke , @PuneetEsri , I believe this is ready for review again. Whilst it's not being merged into master and made public. Thanks.

XuemingRocks:
@TADraeseke @PuneetEsri a friendly reminder to review @LDuncAndroid's PR.

LDuncAndroid:
Thanks @TADraeseke . I've added a comment.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,281 @@
...
+   * @param authenticationChallenge the authentication challenge to handle
+   * @return the AuthenticationChallengeResponse indicating which action to take
+   */
+  override fun handleChallenge(authenticationChallenge: AuthenticationChallenge?): AuthenticationChallengeResponse {
+    authenticationChallenge?.let { authChallenge ->
+
+      try {
+        // if SharedPreferences has an auth code, we've likely just been through the OAuth flow and now have an auth code
+        // we can use to request a new access token
+        sharedPreferences.authCode?.let {
```
PuneetEsri: This is not needed. There is major bug in this sample, that reloads the Portal in onResume() after the user has successfully provided the credential, which causes handleChallenge to be called without finishing the previous OAuth challenge. HandleChallenge() should be called only once per AuthenticationChallenge. This Sample here calls it every time a redirect happens back to this activity.

Also saving auth code in sharedPrefrences is a security flaw and is totally unnecessary for a successful OAuth login workflow.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
PuneetEsri: We have separate workflows for dishing out the challenge in the Browser via the DefaultChallangeHandler and using the WebView. So that we can handle the OAuth challenge from the point we get the Authorization code back from the browser. In the case of DefaultChallengeHandler, the user will have to specify the `DefaultOAuthIntentReceiver` activity so that the ArcGIS api recieves the authorization code and can request OAuth credential. In the case of WebView the control goes back to the shouldOverrideUrlLoading() where they can request OAuthCredential

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,281 @@
...
+          // use the authorization code to get a new access token by executing an OAuthTokenCredentialRequest
+          val request = OAuthTokenCredentialRequest(
```
PuneetEsri: This should be happening in WebView. shouldOverrideUrlLoading()

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: I'd wouldn't call it a major bug, moving that line out of `onResume()` and into `onCreate()` may fix it.

In what way is saving the auth code in `sharedPreferences` a security flaw? It has to be persisted somewhere to be used and since it's cleared after it's used I see no issue with it. Given the fact that the auth code is temporary and the auth code is useless without the secret, I don't see an issue with it at all.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: Did we not agree to not use a `WebView` as 1st choice?

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: We aren't using `DefaultChallengeHandler`, this sample is for a custom Challenge Handler.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
PuneetEsri: >I'd wouldn't call it a major bug, moving that line out of onResume() and into onCreate() may fix it.          

From what I see it will not fix it. 

The flow of the Authentication should be, 
- the user tries to load a Portal, gets challenged with an `OAUTH_CREDENTIAL_CHALLENGE`  in handleChallange(), 
- the app redirects user to the OAuth authentication Portal where the user provides credential, 
- the authenticating portal redirects back to the App with auth code,
-  the App requests for OAuthTokenCredential using the authCode. 

What this sample does is,
-  the user tries to load a Portal, gets challenged with an `OAUTH_CREDENTIAL_CHALLENGE`  in handleChallange(), 
- checks if authcode present in sharedpreferences, if not, then the app redirects user to the OAuth authentication Portal where the user provides credential, 
- the authenticating portal redirects back to the App with auth code,
- the app caches the authcode in SharedPreferences, 
- Loads Portal again, 
- gets challenged with an `OAUTH_CREDENTIAL_CHALLENGE`  in handleChallange() checks sharedpreference for authcode, if present the App requests for OAuthTokenCredential using the authCode. 

So the sample relies on a hack to reload the portal so it can handle the OAuth flow in the HandleChallenge(). We cannot promote this pattern to our users, to tell them that they have to call Load() on a portal twice to complete a successful OAuth authentication.

>In what way is saving the auth code in sharedPreferences a security flaw? It has to be persisted somewhere to be used and since it's cleared after it's used I see no issue with it. Given the fact that the auth code is temporary and the auth code is useless without the secret, I don't see an issue with it at all.     

Persisting any code related to authentication on the client side should be avoided as much as possible cause if you do you are leaving room for vulnerability. You as an app developer might not see a problem with persisting the auth code, but as an api dev we have to push for the right practices/patterns that are secure. 

>It has to be persisted somewhere to be used and since it's cleared after it's used I see no issue with it.       

The only reason why you are persisting the authcode is because you are loading the portal twice to handle the logic in HandleChallenge(). Which is not the right pattern to promote to the users anyway. It is not a requirement to persist the authcode to be able to use it.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
PuneetEsri: Yes, it is not. But the whole purpose of doing this is to show a user how they can do it if they want to do it. If the user decides to use the WebBrowser for the challenge then we should be telling them to use the defaultChallengeHandler. 
Besides, the sample currently relies on loading Portal twice so you might not see my point here.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: In order to perform the `OAuthTokenCredentialRequest` when coming back from after the user has entered their credentials we'd need to block using the `CountdownLatch` as you have in the sample you'd provided. Without blocking the thread we need to attempt to load the Portal to fire the `handleChallenge` call. As an app developer, we regularly rely upon APIs to make callbacks in an asynchronous manner to not block any of the threads and allow the UI experience to continue as designed. If we're manually blocking a thread we have not created app side, I wouldn't like to promote this behaviour.

I have not called `load()` on the Portal, if the `MapView` does this for me, it's likely in `mapView.resume()`, I'd expect this behaviour as we should be calling this method during `onResume()` of the app.

I can remove the persisting to `SharedPreferences` and rely just on the `Intent` having the auth token, but again, this is persisted in runtime in the `Activity` `Intent`. There's no getting around the fact that the auth token must persist somewhere, either in runtime memory or elsewhere dependant on the workflow. A leaked auth token with the correct security principals applied to it server side will only be valid for a short period of time. Facebook is 10 minutes if I recall correctly.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: https://devtopia.esri.com/runtime/java/issues/426#issuecomment-1703139 referring back to the decisions we made

LDuncAndroid:
@PuneetEsri I have moved the setting of the map to `onCreate()` to prevent Portal reloading and I've removed the usages of `SharedPreferences` and instead I'm relying upon the `Intent` in the `Activity` to provide the auth code. Can you re-review please?

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
PuneetEsri: > I have not called load() on the Portal, if the MapView does this for me, it's likely in mapView.resume(), I'd expect this behaviour as we should be calling this method during onResume() of the app.   

This is the call that loads the map 
`mapView.map = ArcGISMap(portalItem)` 

```Sets the ArcGISMap to display in this MapView. By default, the MapView does not contain an ArcGISMap, and one must be set by calling setMap, which then asynchronously loads the map by calling Loadable.loadAsync on the ArcGISMap.```
http://android.esri.com:8080/docs/arcgis-android/publicAPIRelease/WebJavadoc/com/esri/arcgisruntime/mapping/view/MapView.html#setMap(com.esri.arcgisruntime.mapping.ArcGISMap)

Moving this from onResume() to onCreate() will not make a difference. You are reloading the `Map` which loads the `Portal` which issues a second challenge in `handleResponse()`.

>As an app developer, we regularly rely upon APIs to make callbacks in an asynchronous manner to not block any of the threads and allow the UI experience to continue as designed. If we're manually blocking a thread we have not created app side, I wouldn't like to promote this behaviour.     

Well In order to not do it you have put in place a hack that `Reloads` the Map and `Portal` leading to additional requests. I don't think it is acceptable to reload the `Portal` when the previous `Challenge` is not completed just to get the control back in `handleChallenge()`. 

>I can remove the persisting to SharedPreferences and rely just on the Intent having the auth token, but again, this is persisted in runtime in the Activity Intent. There's no getting around the fact that the auth token must persist somewhere, either in runtime memory or elsewhere dependant on the workflow. A leaked auth token with the correct security principals applied to it server side will only be valid for a short period of time. Facebook is 10 minutes if I recall correctly.      

Having the authenticationCode in Activity Intent is OK as that is how we get it back from the Browser.

**Comment by PuneetEsri** [comment]
@LDuncAndroid moving the setMap() call to `onCreate()` will not prevent `Portal` reloading. See my comment here https://github.com/Esri/arcgis-runtime-samples-android/pull/590#discussion_r297386344

File: kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt (hunk 7 above)
Comment (LDuncAndroid) [reply kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt]: @PuneetEsri 
> Well In order to not do it you have put in place a hack that Reloads the Map and Portal leading to additional requests. I don't think it is acceptable to reload the Portal when the previous Challenge is not completed just to get the control back in handleChallenge().

How is it a hack if we've returned the `AuthenticationChallengeResponse` in `handleChallenge()`? The previous challenge is completed and ended with `AuthenticationChallengeResponse.Action.CANCEL`.  I believe it's a hack to block the thread. In all my years as an Android App developer I have yet to come across an API that requires a user to block a thread in a callback.

Comment (LDuncAndroid) [reply kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt]: After having a discussion with @gunt0001 , it's been decided to re-implement the lock and block the thread. It was noted that that the thread can remain blocked if the user fails to return to the app.

**Comment by LDuncAndroid** [comment]
@puneet-pdx I have updated the sample to use a CountdownLatch. Can you re-review please?

File: kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt [hunk 9]
```diff
@@ -0,0 +1,268 @@
...
+
+  /**
+   * Generate the Intent that launches the MainActivity with the new auth code
+   *
+   * @param uri instance of Uri generated from URL that OAuth webpage redirects to after successful login
+   */
+  private fun generateAuthIntent(uri: Uri): Intent {
+    return Intent().also {
+      it.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+      it.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
gunt0001: Can you add some comments here explaining what these flags are for? I assume they prevent the MainActivity from being destroyed?

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,268 @@
...
+fun AppCompatActivity.logToUser(logMessage: String) {
+  Log.d(this::class.java.simpleName, logMessage)
+  Toast.makeText(this, logMessage, Toast.LENGTH_LONG).show()
+}
+
+/**
+ * Intent extensions
+ */
+val Intent.authCode: String?
+  get() = this.data?.getQueryParameter("code")
```
gunt0001: missing newline, also in a few other files...

gunt0001:
@puneet-pdx - please take another look.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,268 @@
...
+   *
+   * We release the thread when the Intent is received and it contains an auth code, setting this auth code as a property
+   * on the [MainActivity] class. When the thread is released we check that the auth code isn't null and perform a
+   * OAuthTokenCredentialRequest.
+   *
+   * @param authenticationChallenge the authentication challenge to handle
+   * @return the AuthenticationChallengeResponse indicating which action to take
+   */
+  override fun handleChallenge(authenticationChallenge: AuthenticationChallenge?): AuthenticationChallengeResponse {
+    authenticationChallenge?.let { authChallenge ->
```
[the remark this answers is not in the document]
LDuncAndroid: I think it's alright to leave it as it is to make it clearer to the developer reading it.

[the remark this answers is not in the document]
gunt0001: Coding style guidelines for Kotlin is something we should start to agree on and write up. Anyway, I agree in this case with Luke, explicitly specifying a parameter name makes the code more readable.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
LDuncAndroid: They're the flags for launching the browser `Intent`. It's `singleTop` on the `Activity` in the manifest that prevents the `Activity` being destroyed.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,268 @@
...
+    // setup AuthenticationManager to handle auth challenges
+    AuthenticationManager.setAuthenticationChallengeHandler(this)
+    AuthenticationManager.addOAuthConfiguration(oAuthConfig)
+  }
+
+  override fun onNewIntent(intent: Intent?) {
+    super.onNewIntent(intent)
+
+    // check the intent for an auth code
+    intent?.authCode?.let {
```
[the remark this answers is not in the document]
LDuncAndroid: https://github.com/Esri/arcgis-runtime-samples-android/pull/590#discussion_r300189392

So it's worth having the null check in the `handleChallenge()` function. If the auth code is null after counting down the lock return `AuthenticationChallengeResponse(AuthenticationChallengeResponse.Action.CANCEL, null)`

LDuncAndroid:
Assigning back to @puneet-pdx for review. Thanks.

LDuncAndroid:
@puneet-pdx 
> @LDuncAndroid the sample does not handle the case when the user presses the back button to cancel the authentication when challenged via browser,
> 
> and also does not dismiss the login page when user presses cancel button in the webview.

This commit resolves it: https://github.com/Esri/arcgis-runtime-samples-android/pull/590/commits/a210f3acf3f59dd76553749f6c88a52b7352f800
And this comment describes it: https://github.com/Esri/arcgis-runtime-samples-android/pull/590#discussion_r306192654

If you have another solution, please let me know.

LDuncAndroid:
I've made those changes @puneet-pdx , back to you for review. Thanks.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    // enabled to allow javascript to run on auth webpage.
+    webView.settings.javaScriptEnabled = true
+    webView.visibility = View.VISIBLE
+  }
+
+  /**
+   * Generate the Intent that launches the MainActivity with the new auth code
+   *
+   * @param uri instance of Uri generated from URL that OAuth webpage redirects to after successful login
```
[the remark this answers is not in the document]
puneet-pdx: > @param uri instance of Uri generated from URL that OAuth webpage redirects to after successful login

This statement is not true. I see this method being called in the case when the user presses cancel on the login page and this is the uri param in that case.

`my-ags-app://auth?error=access_denied&error_description=The%20user%20denied%20your%20request`

LDuncAndroid:
@puneet-pdx 
> That commit only addresses the scenario when the user presses the cancel button on the login page. It does not address the scenario when the user presses the back button on the phone and gets returned back to the MainActivity.
> In my testing I found that when the user presses the back button, onNewIntent() is not called. Only OnResume() is called and the authCountDownLatch is not released. Is this not what you see ? Am I missing something ?

OK, I've moved the counting down of the latch to `onResume()`.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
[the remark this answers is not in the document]
puneet-pdx: @LDuncAndroid this ☝️ comment is still not addressed.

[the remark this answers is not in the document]
LDuncAndroid: @puneet-pdx addressed now.

kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,276 @@
...
+import com.esri.arcgisruntime.security.AuthenticationManager
+import com.esri.arcgisruntime.security.OAuthConfiguration
+import com.esri.arcgisruntime.security.OAuthTokenCredentialRequest
+import kotlinx.android.synthetic.main.activity_main.*
+import java.util.concurrent.CountDownLatch
+
+/**
+ * This sample demonstrates how to authenticate with ArcGIS Online (or your own portal) using OAuth2 to access secured
+ * resources (such as private web maps or layers). Accessing secured items requires a login on the portal that hosts them
+ * (an ArcGIS Online account, for example). This sample utilizes Android WebView to show the OAuth sign-in page in a dialog.
```
puneet-pdx: update the last line to something like....

This sample utilizes custom AuthenticationChallengeHandler where user handles the prompting of Authentication Challenges themselves compared to the DefaultAuthenticationChallengeHandle http://android.esri.com:8080/docs/arcgis-android/publicAPIRelease/WebJavadoc/com/esri/arcgisruntime/security/DefaultAuthenticationChallengeHandler.html where the ArcGISRuntime handles prompting of the Authentication Challenges.
This sample utilizes the system installed web browser to prompt the user with an OAUTH_CREDENTIAL_CHALLENGE 
http://android.esri.com:8080/docs/arcgis-android/publicAPIRelease/WebJavadoc/com/esri/arcgisruntime/security/AuthenticationChallenge.Type.html#OAUTH_CREDENTIAL_CHALLENGE, and include use of the WebView as a fallback mechanism if the user does not have a web browser installed, only to allow you to use this sample in all circumstances. We do not recommend using WebView as part of your OAuth flow as per IETF's recommendations:

https://tools.ietf.org/html/rfc8252

LDuncAndroid:
Thanks for the doc @puneet-pdx . I've updated so ready for re-review.

LDuncAndroid:
Sorry @puneet-pdx . That's been updated.

LDuncAndroid:
Thanks for the review @puneet-pdx .

@TADraeseke could you take a final look at this and merge if it's OK? Thanks.

puneet-pdx:
@LDuncAndroid just to clarify, since we are not releasing this publicly this will not be merged, right?

LDuncAndroid:
> @LDuncAndroid just to clarify, since we are not releasing this publicly this will not be merged, right?

Sorry, right enough @puneet-pdx . Forget that @TADraeseke .

@XuemingRocks Ready to handover to you.

XuemingRocks:
Thanks @LDuncAndroid @puneet-pdx @TADraeseke @gunt0001 for creating such a solid sample. 

As we plan to keep this sample in a branch, closing this PR and notifying our users ...

PR #602, opened 2019-03-22
Luk10114/1314 terrain exag bug

LDuncAndroid:
https://devtopia.esri.com/runtime/common-samples/issues/1314#issuecomment-1590923


java/terrain-exaggeration/src/main/java/com/esri/arcgisruntime/sample/terrainexaggeration/MainActivity.java
```diff
@@ -72,7 +73,7 @@ protected void onCreate(Bundle savedInstanceState) {
       }
 
       @Override public void onStopTrackingTouch(SeekBar seekBar) {
-
+        seekBar.setProgress(Math.max(1, seekBar.getProgress()));
```
TADraeseke: maybe just add a comment like 

`// disallow any progress value below 1`

the only reason I suggest it is because I had to look up what Math.max(int, int) does :P

PR #603, opened 2019-03-27
Trev8939/fix sample names sample viewer

TADraeseke:
Manifest name fixes for sample viewer (sample viewer only reads names from activity as most of the application stuff gets stripped out in the process of converting from app to libraries). While doing it I also formatted them and put in OpenGL ES versions where missing.


LDuncAndroid:
Had a skim over and can't see anything wrong with it. LGTM 👍

PR #607, opened 2019-03-29
Trev8939/display wfs

TADraeseke:
Layer takes 3-4 seconds to load and sometimes doesn't load at all...further investigation required in java/android API, but shouldn't stop sample from going in.

Thanks for the review @LDuncAndroid!


TADraeseke:
We've sorted out a fix now for our various woes. @LDuncAndroid - Please review when you have a chance. Thanks!

java/display-wfs-layer/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+
+}
+
+dependencies {
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
```
LDuncAndroid: Tiny thing but we don't need Constraint Layout dep

PR #608, opened 2019-03-29
Trev8939/kml tour

TADraeseke:
Another one for your review when you have time. Thanks @LDuncAndroid!


java/play-a-kml-tour/README.md
```diff
@@ -0,0 +1,76 @@
...
+   the first KML tour. Once a tour is found, provide it to the KML tour
+   controller. 
+4. Enable the buttons to allow the user to play, pause, and reset the
+   tour.
+   
+## Relevant API
+
+* KmlTourController
+* KmlTourController.Tour
+* KmlTourController.Play()
```
LDuncAndroid: Should we be including the methods used here and not just the classes? I haven't seen this in other samples.

java/play-a-kml-tour/README.md
```diff
@@ -0,0 +1,76 @@
...
+through some of Esri's offices.
+
+## Additional information
+
+See
+[Google's documentation](https://developers.google.com/kml/documentation/touring)
+for information about authoring KML tours.
+
+## Tags
+Visualization
```
LDuncAndroid: Should these tags use different cases?

java/play-a-kml-tour/README.md
TADraeseke: Its in the design, so I think we'll just leave it. The alternative is to have KmlTourController and KmlTour only, which maybe isn't so helpful.

java/play-a-kml-tour/README.md
TADraeseke: They're for search tags anyways, so I don't think it overly matters. Since Android sample viewer uses the first tag for the category at the moment, it just helps me remember whats its doing there :P

TADraeseke:
@LDuncAndroid - Thanks for the review! Addressed your comments above.

PR #613, opened 2019-04-05
Trev8939/display wfs

TADraeseke:


TADraeseke:
@LDuncAndroid - Small fix to remove SRs from point creation thanks to bug fix.

PR #615, opened 2019-04-11
browseWFSfix

TADraeseke:
calling getGeometryType() on the table in getRandomRendererForTable as returning UNKNOWN and so a SimpleLineSymbol was being applied to every renderer


java/browse-wfs-layers/src/main/java/com/esri/arcgisruntime/sample/browsewfslayers/MainActivity.java
```diff
@@ -136,8 +136,8 @@ private void updateMap(WfsLayerInfo wfsLayerInfo) {
     // create a feature layer from the table
     FeatureLayer featureLayer = new FeatureLayer(featureTable);
 
-    // set a renderer for the table
-    featureLayer.setRenderer(getRandomRendererForTable(featureTable));
+    // set a renderer to the table once is loaded, since the renderer is chosen based on the table's geometry type
```
LDuncAndroid: I knew there was a reason for waiting for the table, I must've introduced the regression when refactoring last time.

PR #635, opened 2019-07-24
Trev8939/list kml contents

TADraeseke:
@LDuncAndroid - Can you review as per normal.
@ChrisFrenchPDX - Can you check that I've done the metadata.json file correctly?

Thanks!


java/list-kml-contents/README.md
```diff
@@ -0,0 +1,49 @@
...
+* KmlFolder
+* KmlGroundOverlay
+* KmlLayer
+* KmlNetworkLink
+* KmlNode
+* KmlPlacemark
+* KmlScreenOverlay
+
+## Offline Data
+1. Download the data from [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=da301cb122874d5497f8a8f6c81eb36e).
```
LDuncAndroid: This data requires access to the Samples org.

java/list-kml-contents/src/main/java/com/esri/arcgisruntime/sample/listkmlcontents/MainActivity.java
```diff
@@ -0,0 +1,310 @@
...
+
+public class MainActivity extends AppCompatActivity implements KmlNodeAdapter.OnItemClickListener {
+
+  private static final String TAG = MainActivity.class.getSimpleName();
+
+  private TextView mBreadcrumbTextView;
+  private RecyclerView mRecyclerView;
+  private List<KmlNode> mKmlNodeList;
+  private List<String> mKmlNodeNames;
+  private List<BitmapDrawable> mKmlNodeUxIcons;
```
LDuncAndroid: Since you're rendering the list items in an adapter, it may be better to do all the rendering in there. And instead of using a list of `BitmapDrawable` (as they could be quite resource intensive, bitmaps usually are) you might find using an `LRUCache` is more performant. https://developer.android.com/reference/android/util/LruCache

However, if you'd rather focus on usage of our API, ignore this as it's an implementation detail in Android.

java/list-kml-contents/README.md
```diff
@@ -0,0 +1,49 @@
...
+1. Extract the contents of the downloaded zip file to disk.
```
LDuncAndroid: I just got a `.kmz` file so no need to unzip.

TADraeseke: oops yeah...

java/list-kml-contents/src/main/java/com/esri/arcgisruntime/sample/listkmlcontents/MainActivity.java
TADraeseke: Yeah the obvious thing to do would be to pass a list of kml nodes to the adapter, but my worry was that I'd then be moving the relevant API calls to the adapter instead of in the main activity

TADraeseke:
Thanks @LDuncAndroid - Made those fixes and did some linting. Ready for re-review :)

PR #636, opened 2019-07-24
identifyKml metadata.json

TADraeseke:
One more to check when you have a sec @ChrisFrenchPDX

### DIFF
--- a/java/identify-kml-features/README.metadata.json
+++ b/java/identify-kml-features/README.metadata.json
@@ -1,0 +1,29 @@
+{
+    "category": "Search and Query",
+    "description": "Show a balloon popup with formatted content for tapped KML features. KML supports defining HTML for popups.",
+    "ignore": false,
+    "images": [
+        "identify-kml-features.png"
+    ],
+    "keywords": [
+        "Weather",
+	"NOAA",
+	"NWS",
+	"KML",
+	"KMZ", 
+	"OGC",
+	"Keyhole"
+    ],
+    "redirect_from": "/android/latest/sample-code/identify-graphics.htm",
+    "relevant_apis": [
+        "KmlLayer",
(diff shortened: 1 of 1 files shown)


java/identify-kml-features/README.metadata.json
```diff
@@ -0,0 +1,29 @@
...
+	"KML",
+	"KMZ", 
+	"OGC",
+	"Keyhole"
+    ],
+    "redirect_from": "",
+    "relevant_apis": [
+        "KmlLayer",
+        "KmlPlacemark",
+        "KmlPlacemark.getBalloonContent()",
```
ChrisFrenchPDX: This is just a side question, out of curiosity - Is there a reason for the extra syntax in the type/method names? Eventually, I was thinking of adding a feature to implement a way to generate links to API reference pages from these `relevant_apis` field, but I would need to strip out the parentheses and ellipses to use the class name to autogenerate the url, but maybe this is more difficult than I think.

For example, for Android, the `relevant_apis` would need to include the entire namespace of each type, e.g. 

`com.esri.arcgisruntime.mapping.ArcGISMap` has the url

```url
https://developers.arcgis.com/android/latest/api-reference/reference/com/esri/arcgisruntime/mapping/ArcGISMap.html
```

which has the schema

```url
https://developers.arcgis.com/android/latest/api-reference/reference/<namespace slug>/**/<namespace_slug>/<type name>.html
```

so, assuming the namespace prefix `com.esri.arcgisruntime` is constant for all the android apis, to add this api auto-linking feature, the `relevant_apis` field would need to include `mapping.ArcGISMap` and not `ArcGISMap`, unless it's always the case that the entire namespace of a type can be deterministicly inferred from the type name alone.

PR #644, opened 2019-08-15
Xueming/update iwa sample

TADraeseke:
@XuemingRocks has made these updates because of a change in cancellation reporting by the AuthenticationChallengeResponse.


java/integrated-windows-authentication/src/main/java/com/esri/arcgisruntime/sample/integratedwindowsauthentication/MainActivity.java
```diff
@@ -229,14 +228,6 @@ public AuthenticationChallengeResponse handleChallenge(AuthenticationChallenge a
             authenticationChallenge);
       }
 
-      int maxAttempts = 5;
```
TADraeseke: Is this handled within the challenge handler now? Or why has this been removed?

XuemingRocks: The SDK always handles max attempts. When the number of failed attempts exceeds the number of max attempts, the SDK throws the original exception that causes the challenge which could be a `HttpResponseException` with message saying `Unauthorized access` when dealing with IWA.  The number of max attempts is controlled by `RequestConfiguration`. And the default value is 5. If this custom handler wanted to handle exceeding max attempts with `AuthenticationChallengeResponse.Action.CANCEL` which now throws CancellationException and message saying "User cancelled authentication challenge". It may not work as expected.  When the max attempts defined in the custom handler is not greater than 5, it throws the CancellationException. If it is greater than 5, the SDK throws the original exception when the number of failed attempts is equal to 5. So when users change this sample to set the max attempt in the custom handler to a number greater than 5, it won't work as expected. And it may cause confusion by throwing different exceptions when the number of max attempts are different.  

We can add a few more lines to set the number of max attempts through `RequestConfiguration`. However, it makes the sample a bit complicated and loses focus. I think we should keep this sample simple and clean. So, I removed that block. 

Please let me if you have other thoughts.

TADraeseke:
All looks goof to me, thanks @XuemingRocks!

PR #647, opened 2019-09-11
Trev8939/display tabletop map ar

TADraeseke:
Hi Luke,

Please pick this apart--this was very much just me:
A) Trying to mimic Nathan's initial iOS implementation
B) Me having a bit of a nosey around some of ArCore.

Things to look out for:
- Am I using calls from ArCore that I should instead by accessing through the toolkit?
  - Have I handled touches correctly?
  - Is there an easier way to access the plane from the hitResult to call `getExtentX()`?

- Does the math make plausible sense around `setTranslationFactor()`?

- Do I need to make those calls to setTracking(true) and planeRenderer enabled/visible?

- By waiting for a hit result of type `plane` in this sample, am I excluding phones which can't run ArCore from having some sort of operation just off sensors?


LDuncAndroid:
Sorry for taking so long to get to this review, I was fixing issues in the Toolkit :)

TADraeseke:
BEFORE RELEASE:
- Update README links
- Change aar links to bin tray

PR #649, opened 2019-09-13
Trev8939/navigate route

TADraeseke:
Navigate a route sample: https://devtopia.esri.com/runtime/common-samples/issues/1597

Any chance you could review this for me @eri9600? Thanks!


java/navigate-route/src/main/java/com/esri/arcgisruntime/sample/navigateroute/MainActivity.java
```diff
@@ -0,0 +1,280 @@
...
+        RouteParameters routeParameters = routeParametersFuture.get();
+        routeParameters.setStops(getStops());
+        routeParameters.setReturnDirections(true);
+        routeParameters.setReturnStops(true);
+        routeParameters.setReturnRoutes(true);
+        ListenableFuture<RouteResult> routeResultFuture = routeTask.solveRouteAsync(routeParameters);
+        routeParametersFuture.addDoneListener(() -> {
+          try {
+            // get the route geometry from the route result
+            RouteResult routeResult = routeResultFuture.get();
```
eri9600: It seems that you are doing things twice: getting the geometry, creating the graphic for the route, and setting it to the graphics overlay, as this is done here and inside the function `startNavigation`. `startNavigation` gets called in line 127 so is there any benefit for doing these things twice?

TADraeseke: Ahh I think I see what you mean, but one is inside the onclick listener and one is not. So basically:

- startNavigation is called once, once a route as been calculated
- startNavigation is called again whenever the "Navigate Route" button is pressed

eri9600: @TADraeseke Thanks for clarifying.

TADraeseke:
Thanks @eri9600 -- Sorry for delay. Added messages about stops to clarify. Thanks for the review!

PR #673, opened 2019-11-25
Trev8939/custom dictionary style

TADraeseke:
I'll start cycling round Java/Android home team folk with follow on samples, but for now--any thoughts and concerns welcomed. Some new patterns in this sample:

- Use getExternalFilesDir(...) to be more Android 10 compliant
- Try to use kotlin scoping functions in good faith


kotlin/custom-dictionary-style/src/main/java/com/esri/arcgisruntime/sample/customdictionarystyle/MainActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+            getExternalFilesDir(getString(R.string.restaurant_stylx_path))?.path))
+        // once the feature layer has loaded
+        this.addDoneLoadingListener {
+          // set the map view's viewpoint to the feature layer extent
+          mapView.setViewpointAsync(Viewpoint(this.fullExtent))
+        }
+        // add the the feature layer to the map's operational layers
+        operationalLayers.add(this)
+      }
+    }
```
tschie: I would prefer:
```kotlin
    // create a new map with a streets basemap and set it to the map view
    val map = ArcGISMap(Basemap.createStreetsVector())
    mapView.map = map

    // create a feature layer from a service feature table
    val featureTable = ServiceFeatureTable(getString(R.string.restaurants_url))
    val featureLayer = FeatureLayer(featureTable).apply {
      // use a custom style to create a dictionary renderer and set it to the feature layer renderer
      val dictionaryStyle = DictionarySymbolStyle.createFromFile(getExternalFilesDir(getString(R.string.restaurant_stylx_path))?.path)
      renderer = DictionaryRenderer(dictionaryStyle)
    }

    // set the map view's viewpoint to the feature layer extent when loaded
    featureLayer.addDoneLoadingListener { 
      mapView.setViewpointAsync(Viewpoint(featureLayer.fullExtent))
    }

    // add the the feature layer to the map's operational layers
    map.operationalLayers.add(featureLayer)
```

Use `apply` only in cases like this:
```kotlin
val myObject = MyObject("somethingA")
myObject.somethingC = ""
myObject.somethingF = ""
```
```kotlin
val myObject = MyObject("somethingA").apply {
  somethingC = ""
  somethingF = ""
}
```
It's just so that you don't have the redundant `myObject`. It's almost as if you're creating an overload constructor for `MyObject` that only takes the arguments you want to set. It replaces the builder pattern if you're familiar with that. Here, it would be nice if there was a FeatureLayer constructor that took a feature table and a renderer. The `apply` here nicely shows that you "built" a FeatureLayer using a required feature table with an "applied" renderer.

The mechanism that powers `with` is receiver type functions. One major use for those are DSLs. So you'll often do things with `with` that you'd maybe do with a DSL. For example, a SQL string builder:
```kotlin
val query = with (select("*")) {
   val fromStatement = if (something) {
      // Select.from
      from(MyTable)
   } else {
      from(AnotherTable)
  }

  with (fromStatement) {
    // From.where
    where(...)
  }
}
```

You can do a bunch of things *with* the scope of `select` to get your `from` and then do something *with* `from` and so on in a chaining manner. I wouldn't expect you to use it in samples much. Maybe if you were building a tree structure or a complex query based on a bunch of conditionals.

TADraeseke: Thanks for this @tschie - This is super useful for me trying to get the patterns of these uses right!

TADraeseke: Changed most things to fix your proposed patterns. I've stuck with doing the "map" related work all in the same place, because it still feels like building up the contents of the map with:

```
    // create a new map with a streets basemap and set it to the map view
    mapView.map =  ArcGISMap(Basemap.createStreetsVector()).apply {
      // add the the feature layer to the map's operational layers
      operationalLayers.add(featureLayer)
    }
```

kotlin/custom-dictionary-style/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    // arcgis-android & appcompat lib dependencies from rootProject build.gradle
+    implementation "org.jetbrains.anko:anko-commons:$ankoVersion"
```
gunt0001: As discussed, we cannot use this dependency as it has been deprecated.

TADraeseke:
Thanks @gunt0001 - Removed anko, and updated `getExternalFilesDir(...)` to the new pattern. Ready for re-review.

kotlin/custom-dictionary-style/src/main/java/com/esri/arcgisruntime/sample/customdictionarystyle/MainActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+      // use a custom style to create a dictionary renderer and set it to the feature layer renderer
+      renderer = DictionaryRenderer(
+        DictionarySymbolStyle.createFromFile(
+          getExternalFilesDir(null)?.path + getString(R.string.restaurant_stylx_path)
+        )
+      )
+    }
+
+    // set the map view's viewpoint to the feature layer extent when loaded
+    featureLayer.addDoneLoadingListener {
```
gunt0001: you could add this to the `FeatureLayer(featureTable).apply { ...}` block above

TADraeseke: That'd be my intuition too, but I'm trying to keep to the apply block as a sort of extended constructor as per @tschie's comment above.

PR #685, opened 2020-01-23
Add multidex dependency

Rachael-E:
Android has a limit of 64k methods in a single app (including all libraries on which it depends). Some samples have started to hit that number and Android's way around it (pre API 21) is with multidex. This PR adds multidex dependency at the project level build.gradle script to catch any samples where this might occur. We will remove this when our min API is 21.


Rachael-E:
thanks @TADraeseke, added in for Java now!

PR #688, opened 2020-02-01
Sor10874/198 spatial operations kotlin

sorenoid:
hi all, another PR now that I've gotten Trevor's feedback on the first. thanks!


kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,209 @@
...
+      add(Point(-13160.0, 6710100.0))
+      add(Point(-13300.0, 6710500.0))
+      add(Point(-13760.0, 6710730.0))
+      add(Point(-14660.0, 6710000.0))
+      add(Point(-13960.0, 6709400.0))
+      inputPolygon1 = Polygon(this)
+    }
+
+    // create and add a blue graphic to show input polygon 1
+    SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, -0x66ffff34, lineSymbol).let {
```
TADraeseke: Let's go with `Color.BLUE` instead, just to make it a bit more readable

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,209 @@
...
+  private var differenceMenuItem: MenuItem? = null
+  private var symmetricDifferenceMenuItem: MenuItem? = null
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_main)
+
+    // create ArcGISMap with topographic basemap
+    ArcGISMap(Basemap.createLightGrayCanvas()).let {
+      mapView.map = it
```
TADraeseke: Use a mapView.apply to do alot of this work as part of its instantiation

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,209 @@
...
+      inputPolygon1 = Polygon(this)
+    }
+
+    // create and add a blue graphic to show input polygon 1
+    SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, -0x66ffff34, lineSymbol).let {
+      inputGeometryOverlay.getGraphics().add(Graphic(inputPolygon1, it))
+    }
+
+    // outer ring
+    val outerRing = Part(PointCollection(SpatialReferences.getWebMercator()).also {
```
TADraeseke: Also is generally used to do stuff in addition to the creation of an object like

.also {
Log.d(TAG, "This is called: " + it.name
}

Let's use apply. { ... } here

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,209 @@
...
+import com.esri.arcgisruntime.symbology.SimpleLineSymbol
+import kotlinx.android.synthetic.main.activity_main.*
+
+class MainActivity : AppCompatActivity() {
+
+  private val inputGeometryOverlay: GraphicsOverlay by lazy { GraphicsOverlay() }
+  private val resultGeometryOverlay: GraphicsOverlay by lazy { GraphicsOverlay() }
+
+  // simple black (0xFF000000) line symbol for outlines
+  private val lineSymbol = SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, -0x1000000, 1f)
```
TADraeseke: Use `Color.BLACK` for readability, where possible.

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_main)
+
+    mapView.apply {
+      // create ArcGISMap with topographic basemap
+      map = ArcGISMap(Basemap.createLightGrayCanvas())
+
+      // create graphics overlays to show the inputs and results of the spatial operation
+      graphicsOverlays.add(inputGeometryOverlay)
```
Rachael-E: I think we could name the graphicsOverlays more clearly for our users: how about

`inputGeometryGraphicsOverlay`
`resultGeometryGraphicsOverlay`

I realise it's a little longer but it helps a user keep track of what type of object the variable being used is, i.e. we don't have an API of `GeometryOverlay` which currently I think name suggests.

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+      graphicsOverlays.add(resultGeometryOverlay)
+
+      // create input polygons and add graphics to display these polygons in an overlay
+      createPolygons()
```
Rachael-E: I'm still super new to Kotlin, but it's my understanding that when using the `apply` scoping function, we shouldn't be adding methods whose logic doesn't relate directly to the object using apply. In this instance `createPolygons` is creating polygons, which isn't directly linked to the set up of the map view object. I'm wondering would that be better called outside of the `mapView.apply` block, along with the following envelope and `setViewpointGeometryAsync` set up and am raising this as a discussion point. What do you think @TADraeseke?

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+
+    // create and add a green graphic to show input polygon 2
+    SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, -0x66ff6700, lineSymbol).let {
+      inputGeometryOverlay.graphics.add(Graphic(inputPolygon2, it))
+    }
+
+  }
+
+  override fun onCreateOptionsMenu(menu: Menu?): Boolean {
+    // Inflate the menu; this adds items to the action bar if it is present.
```
Rachael-E: this is super nit picky, but just for consistency in style across samples, all comments should follow one style. e.g. the majority of comments in this sample start off lower case, whereas there are some in this method which start off sentence case.

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+import com.esri.arcgisruntime.symbology.SimpleFillSymbol
+import com.esri.arcgisruntime.symbology.SimpleLineSymbol
+import kotlinx.android.synthetic.main.activity_main.*
+
+class MainActivity : AppCompatActivity() {
+
+  private val inputGeometryOverlay: GraphicsOverlay by lazy { GraphicsOverlay() }
+  private val resultGeometryOverlay: GraphicsOverlay by lazy { GraphicsOverlay() }
+
+  // simple black (0xFF000000) line symbol for outlines
```
TADraeseke: can drop the hex code

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+  private val lineSymbol = SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.BLACK, 1f)
+  private val resultFillSymbol = SimpleFillSymbol(
+    SimpleFillSymbol.Style.SOLID, -0x16e0e1, lineSymbol
```
TADraeseke: Let's use android Color.CYAN or something similar for readability

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+  )
+  private var inputPolygon1: Polygon? = null
```
TADraeseke: these could probably be `lateinit` since you're committing to initiating these in onCreate

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+        return true
+      }
+
+      R.id.action_difference -> {
+        differenceMenuItem?.isChecked = true
+        // note that the difference method gives different results depending on the order of input geometries
+        showGeometry(GeometryEngine.difference(inputPolygon1, inputPolygon2))
+        return true
+      }
+
```
TADraeseke: can drop the white space lines between options

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+      val envelope = GeometryEngine.union(inputPolygon1, inputPolygon2).extent
+      setViewpointGeometryAsync(envelope, 20.0)
+    }
+
+  }
+
+  private fun showGeometry(resultGeometry: Geometry) {
+    // add a graphic from the result geometry, showing result in red (0xFFE91F1F)
+    Graphic(resultGeometry, resultFillSymbol).let {
+      resultGeometryOverlay.graphics.add(it)
```
TADraeseke: This pattern might be a bit hard to read...usually the stuff in the scoping function is stuff happening to the receiver rather than the receiver being added to something else. Maybe move towards something like:

```
    val graphic = Graphic(resultGeometry, resultFillSymbol).also {
      it.isSelected = true
    }
    resultGeometryOverlay.graphics.add(graphic)
```

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
sorenoid: yes I am a bit shaky on what can and cannot be done in apply{}. It seems it is intended for initialization, but I am a bit lost with the merits of calling something like createPolygons() inside of apply{} versus the merits of, say, chaining a run {} onto the end of it, or just adding a few more lines of code. Most advice I find is to use the scope functions to reduce the amount of code needed, so that is why I called createPolygons() in apply {}. At first I was worried that the newly constructed object might not be fully intialized in apply{}, but this seems to not be the case (can someone confirm this?) and that the choice of scope function is a question of convention or design pattern.

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,217 @@
...
+      add(Point(-13160.0, 6710100.0))
+      add(Point(-13300.0, 6710500.0))
+      add(Point(-13760.0, 6710730.0))
+      add(Point(-14660.0, 6710000.0))
+      add(Point(-13960.0, 6709400.0))
+      inputPolygon1 = Polygon(this)
+    }
+
+    // create and add a blue graphic to show input polygon 1
+    SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, Color.BLUE, lineSymbol).let {
```
TADraeseke: I might break this up too

kotlin/spatial-operations/README.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+        "Edit and Manage Data",
+        "GeometryEngine.difference",
+        "GeometryEngine.intersection",
+        "GeometryEngine.symmetricDifference",
+        "GeometryEngine.union",
+        "Graphic",
+        "GraphicsOverlay",
+        "MapView.setViewpointGeometryWithPaddingAsync"
+    ],
+    "redirect_from": "",
```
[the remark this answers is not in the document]
sorenoid: I noticed that some other kotlin samples have blank redirect_from fields.  But looking more closely, I see that most do have the field filled out.

kotlin/spatial-operations/README.md
```diff
@@ -0,0 +1,22 @@
+# Spatial Operations
```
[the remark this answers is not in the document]
sorenoid: I think I have done this. checked with @mikewilburn about the status of the readme in the common samples design repo, and then simply copied the README.md from the common-samples design repo. I checked the API (it is the same), although there is one method that wasn't in the new README.md or metadata, so I added it to the API in the md and metadata files, as well as the keywords in the metadata.  The method is "MapView.setViewpointGeometryWithPaddingAsync"  and is indeed called. Please let me know if this is ok.

kotlin/spatial-operations/src/main/java/com/esri/arcgisruntime/sample/spatialoperations/MainActivity.kt
sorenoid: i used RED to match the other platforms

kotlin/perform-spatial-operations/README.metadata.json
```diff
@@ -3,30 +3,37 @@
...
-        "Edit and Manage Data",
+	"analysis",
+	"combine",
+	"difference",
+	"geometry",
+	"intersection",
+	"merge",
+	"polygon",
+	"union",
         "GeometryEngine.difference",
```
Rachael-E: this and the following should be removed as they come under the relevant api section rather than keywords (GeometryEngine.difference through to GraphicsOverlay)

kotlin/perform-spatial-operations/README.metadata.json
```diff
@@ -3,30 +3,37 @@
...
         "GeometryEngine.union",
         "Graphic",
         "GraphicsOverlay",
-        "MapView.setViewpointGeometryWithPaddingAsync"
     ],
-    "redirect_from": "",
+    "redirect_from": "/android/latest/sample-code/spatial-operations.htm",
     "relevant_apis": [
-        "GeometryEngine.difference",
+	"Geometry",
```
Rachael-E: let's keep the indentation consistent here and in line

kotlin/perform-spatial-operations/src/main/java/com/esri/arcgisruntime/sample/performspatialoperations/MainActivity.kt
```diff
@@ -131,14 +130,13 @@ class MainActivity : AppCompatActivity() {
     }
 
     // create and add a green graphic to show input polygon 2
-    SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, -0x66ff6700, lineSymbol).let {
-      inputGeometryOverlay.graphics.add(Graphic(inputPolygon2, it))
-    }
+    val greenFill = SimpleFillSymbol(SimpleFillSymbol.Style.SOLID, -0x66ff6700, lineSymbol)
```
Rachael-E: to be consistent with the above `SimpleFillSymbols`, should probably replace this hex code with `Color.GREEN`

sorenoid: I found Color.GREEN a bit hard to look at, but I am not sure what is more important here. I am happy to make the change!

kotlin/perform-spatial-operations/src/main/java/com/esri/arcgisruntime/sample/performspatialoperations/MainActivity.kt
```diff
@@ -0,0 +1,210 @@
...
+      // select the result to highlight it
+      it.isSelected = true
+    }
+
+    resultGeometryGraphicsOverlay.graphics.add(graphic)
+  }
+
+  private fun createPolygons() {
+    // create input polygon 1
+    PointCollection(SpatialReferences.getWebMercator()).run {
```
TADraeseke: maybe change this to an apply like inner and outer ring? `inputPolygon1 = Polygon(PointCollection(SpatialRef)).apply {`

sorenoid: sorry about missing that, should have seen it.

PR #689, opened 2020-02-04
Kotlin sample: export tiles

Rachael-E:
hey @TADraeseke, another kotlin sample for you to take a look at when you can! The biggest changes I made from the Java Android sample, is that rather than the preview window showing a larger mapview area than the one exported, the preview window now shows exactly the map exported as defined by the preview window (it follows the Java SDK workflow in that way). I think this is a little clearer to the user, but please let me know what you think.

Also have the new progress indicator layout: some issues I have with it are I haven't figured out how to chance its indicator from orange to blue (to keep with our colour scheme) and that you can click away from it and the progress disappears (but the job still completes). Not sure if we need to add in an indicator somewhere there is still a download happening in the background. 

One final thing, I'm getting a crash when rotating the device: I haven't figured out why, but it's to do with the geometry of the download area returning null (see code block for where crash occurs). I'm not sure why this should be when rotating the device (if it's just reloading the app, I don't see this crash on start up). 

```                exportTileCacheTask.createDefaultExportTileCacheParametersAsync(downloadArea.geometry, mapView.mapScale, tiledLayer.maxScale)
```


kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,240 @@
...
+    setContentView(R.layout.activity_main)
+
+    // create an ArcGISTiledLayer to use as the basemap
+    val tiledLayer = ArcGISTiledLayer(getString(R.string.world_street_map))
+    val map = ArcGISMap().apply {
+      basemap = Basemap(tiledLayer)
+      minScale = 10000000.0
+    }
+    // create a graphics overlay
+    val graphicsOverlay = GraphicsOverlay()
```
TADraeseke: Could reorder a few of these instantiations to use all the applyz

```
    // create an ArcGISTiledLayer to use as the basemap
    val tiledLayer = ArcGISTiledLayer(getString(R.string.world_street_map))
    val map = ArcGISMap().apply {
      basemap = Basemap(tiledLayer)
      minScale = 10000000.0
    }

    // create a graphic to show a red box around the tiles we want to download
    val downloadArea = Graphic().apply {
      symbol = SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.RED, 2f)
    }
    
    // create a graphics overlay
    val graphicsOverlay = GraphicsOverlay().apply { 
      graphics.add(downloadArea)
    }
    
    mapView.apply {
      // set the map to the map view
      this.map = map
      setViewpoint(Viewpoint(51.5, 0.0, 10000000.0))
      // add the graphics overlay to the map view
      graphicsOverlays.add(graphicsOverlay)
    }
```

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,240 @@
...
+      if (mapView.map.loadStatus == LoadStatus.LOADED) {
+        // upper left corner of the downloaded tile cache area
+        val minScreenPoint = Point(150, 175)
+        // lower right corner of the downloaded tile cache area
+        val maxScreenPoint = Point(
+          mapView.getWidth() - 150,
+          mapView.getHeight() - 250
+        )
+        // convert screen points to map points
+        val minPoint: com.esri.arcgisruntime.geometry.Point? =
```
TADraeseke: I usually try to import the Runtime Point preferentially over the Android Point so that Runtime Point doesn't look so out of place in the code

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,240 @@
...
+          mapView?.screenToLocation(minScreenPoint)
+        val maxPoint: com.esri.arcgisruntime.geometry.Point? =
+          mapView?.screenToLocation(maxScreenPoint)
+        // use the points to define and return an envelope
+        val envelope = Envelope(minPoint, maxPoint)
```
TADraeseke: I'd just inline Envelope here since its not used elsewhere

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    return builder.create()
+  }
+
+  /**
+   * Clear the preview window.
+   */
+  private fun clearPreview() {
+    previewMapView.getChildAt(0).visibility = View.INVISIBLE
+    mapView.bringToFront()
```
TADraeseke: I made the original sample back when I didn't know alot about Android API -- is this call still needed?

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  override fun onResume() {
+    super.onResume()
+    mapView.resume()
+  }
+
+  override fun onPause() {
+    mapView.pause()
+    deleteDirectory(getCacheDir())
```
TADraeseke: maybe just add a comment

// delete app cache when the app loses focus

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,240 @@
...
+
+      parametersFuture.addDoneListener {
+        val parameters: ExportTileCacheParameters = parametersFuture.get()
+        // export tiles to temporary cache on device
+        val exportTileCacheJob =
+          exportTileCacheTask.exportTileCache(parameters, exportTilesDirectory.path + "file.tpk")
+        exportTileCacheJob.start()
+        // create a progress dialog to show export tile progress
+        val dialog = createProgressDialog(exportTileCacheJob).apply {
+          getButton(DialogInterface.BUTTON_NEUTRAL).setTextColor(R.style.progressBarBlue)
```
TADraeseke: Hmm this line gives a null object reference for me when I hit "Export Tiles"

kotlin/export-tiles/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,145 @@
...
+      android:fontFamily="sans-serif-medium"
+      android:text="@string/preview_text"
+      android:textColor="@color/colorPrimary"
+      android:textSize="26sp"
+      app:layout_constraintBottom_toTopOf="@+id/previewMapView"
+      app:layout_constraintHorizontal_bias="0.494"
+      app:layout_constraintLeft_toLeftOf="parent"
+      app:layout_constraintRight_toRightOf="parent"
+      app:layout_constraintTop_toTopOf="parent"
+      tools:layout_constraintLeft_creator="1"
```
TADraeseke: Not sure these creator things are needed (I appreciate those may be leftover from my implementation of the java sample), can you try removing them?

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
Rachael-E: Haha so many applyz! I'm hesitant to use apply for single line instances, purely because it actually adds an extra line of code doing it this way and it doesn't make it quite so readable e.g. the above would mean 6 lines of codes, whereas I think the following is probably more readable and is only 4 lines: 

```
    val graphicsOverlay = GraphicsOverlay()
    val downloadArea = Graphic()
    graphicsOverlay.graphics.add(downloadArea)
    downloadArea.symbol = SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.RED, 2f)
```

TADraeseke: Hmm...I find the 6 lines more readable myself (I think the indents help). It's more lines, but arguably less code. But anyways, if you're keen on the more java-y build pattern I'd at least suggest re-ordering it like:

```
    val downloadArea = Graphic()
    downloadArea.symbol = SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.RED, 2f)

    val graphicsOverlay = GraphicsOverlay()
    graphicsOverlay.graphics.add(downloadArea)
```

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
Rachael-E: yup, `mapView.bringToFront` is still needed otherwise the preview pane gets in the way of the `mapView`: but I realised we can now do without the visibility call on `previewMapView`, thanks for raising this!

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
Rachael-E: it's started doing that for me too: I'll just remove it, it's not important.

kotlin/export-tiles/src/main/res/layout/activity_main.xml
Rachael-E: nice one, I've taken them out and all is well!

Rachael-E:
Thanks @TADraeseke ! Could you take another look please? Still having problems understanding why I'm getting a crash when rotating to landscape. When you have a moment could you take a look please and see if anything obvious jumps out at you as to why and let me know please? Thanks!

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,229 @@
...
+        exportTileCacheTask.createDefaultExportTileCacheParametersAsync(
+          downloadArea.geometry,
+          mapView.mapScale,
+          tiledLayer.maxScale
+        )
+
+      parametersFuture.addDoneListener {
+        val parameters: ExportTileCacheParameters = parametersFuture.get()
+        // export tiles to temporary cache on device
+        val exportTileCacheJob =
```
TADraeseke: I think this is a good opportunity to use a scoping function to hold all the various listeners:

```
 val parameters: ExportTileCacheParameters = parametersFuture.get()
        // export tiles to temporary cache on device
        val exportTileCacheJob =
          exportTileCacheTask.exportTileCache(parameters, exportTilesDirectory.path + "file.tpk")
            .apply {

              // show progress of the export tile cache job on the progress bar
              addProgressChangedListener {
                dialog.progressBar.progress = progress
              }

              addJobDoneListener {
                dialog.dismiss()
                if (status == Job.Status.SUCCEEDED) {
                  val exportedTileCacheResult: TileCache = result
                  showMapPreview(exportedTileCacheResult)
                } else {
                  ("Job did not succeed: " + error.additionalMessage).also {
                    Toast.makeText(this@MainActivity, it, Toast.LENGTH_LONG).show()
                    Log.e(TAG, error.additionalMessage)
                  }


                }

              }
            }
        // create and show a progress dialog to show export tile progress
        dialog = createProgressDialog(exportTileCacheJob)
        dialog.show()
```

or something like that

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,229 @@
...
+      val exportTileCacheTask = ExportTileCacheTask(tiledLayer.uri)
+      val parametersFuture: ListenableFuture<ExportTileCacheParameters> =
+        exportTileCacheTask.createDefaultExportTileCacheParametersAsync(
+          downloadArea.geometry,
+          mapView.mapScale,
+          tiledLayer.maxScale
+        )
+
+      parametersFuture.addDoneListener {
+        val parameters: ExportTileCacheParameters = parametersFuture.get()
```
TADraeseke: let's use try-catch around our future gets

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,229 @@
...
+            Toast.makeText(
+              this, exportTileCacheJob.error.additionalMessage, Toast.LENGTH_LONG
+            ).show()
+          }
+        }
+      }
+    }
+
+    // close the preview window
+    closeButton.setOnClickListener { clearPreview() }
```
TADraeseke: we can use databinding for this:

add `android:onClick="clearPreview"` to the button in xml

and change the private `fun clearPreview()` method to `clearPreview(view: View)`

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,229 @@
...
+    } else if (dir != null && dir.isFile) {
+      dir.delete()
+    } else {
+      false
+    }
+  }
+
+  override fun onResume() {
+    super.onResume()
+    mapView.resume()
```
TADraeseke: will need the same calls to previewMapView

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,229 @@
...
+    val newTiledLayer = ArcGISTiledLayer(result)
+    val previewMap = ArcGISMap(Basemap(newTiledLayer))
+
+    previewMapView.apply {
+      map = previewMap
+      setViewpoint(mapView.getCurrentViewpoint(Viewpoint.Type.CENTER_AND_SCALE))
+      visibility = View.VISIBLE
+      getChildAt(0).visibility = View.VISIBLE
+    }
+    mapPreviewLayout.bringToFront()
```
TADraeseke: I think its probably confusing using z-ordering here. When I wrote the java version of this I didn't have a good understanding of Android UI. Can we just use `visibility=View.VISIBLE` here and invisible elsewhere?

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * Create a progress dialog box for tracking the export tile cache job.
+   *
+   * @param exportTileCacheJob the export tile cache job progress to be tracked
+   * @return an AlertDialog set with the dialog layout view
+   */
+  private fun createProgressDialog(exportTileCacheJob: ExportTileCacheJob): AlertDialog {
+    val builder = AlertDialog.Builder(this@MainActivity)
+    with(builder) {
```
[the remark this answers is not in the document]
Rachael-E: oh cool, thanks for flagging that up!

Rachael-E:
@sorenoid please could you do a second review for this sample? @TADraeseke and I are still working on a way to make the progress bar colours match the rest of the app: other than that work in progress (no code in here yet for that) everything else is ready for the second review. Thank you!

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,251 @@
...
+      view.visibility = View.VISIBLE
+    }
+  }
+
+  /**
+   * Recursively deletes all files in the given directory.
+   *
+   * @param dir to delete
+   */
+  private fun deleteDirectory(dir: File?): Boolean {
```
sorenoid: could just call File.deleteRecursively() here https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/delete-recursively.html

Rachael-E: nice one @sorenoid ! I found a .deleteRecursively attached directly to the cacheDir as well so have used that. Good spot cos it turned out I didn't have the creating of the cacheDir set up in the correct place anyway: I have fixed that now too.

Rachael-E:
@TADraeseke I've implemented those changes we discussed: this should now be working on all devices! could you take another look through please? Really like your addition to the progress bar, thanks for that!

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,263 @@
...
+            }
+        } catch (e: InterruptedException) {
+          Toast.makeText(this, "TileCacheParameters interrupted: " + e.message, Toast.LENGTH_LONG)
+            .show()
+          Log.e(TAG, "TileCacheParameters interrupted: " + e.message)
+        } catch (e: ExecutionException) {
+          Toast.makeText(this, "Error generating parameters: " + e.message, Toast.LENGTH_LONG)
+            .show()
+          Log.e(TAG, "Error generating parameters: " + e.message)
+        }
```
TADraeseke:
```suggestion
          catch (e: Exception) {
          val error = "Error generating tile cache parameters: ${e.message}"
          Toast.makeText(this, error, Toast.LENGTH_LONG).show()
          Log.e(TAG, error)
        } 
```

TADraeseke: Since the difference between interrupted vs Execution exceptions aren't too important. Perhaps its a bit easier to read as above?

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,263 @@
...
+    previewMapView.apply {
+      map = previewMap
+      setViewpoint(mapView.getCurrentViewpoint(Viewpoint.Type.CENTER_AND_SCALE))
+    }
+    // control UI visibility
+    previewMapView.getChildAt(0).visibility = View.VISIBLE
+    show(closeButton, dimBackground, preview_text_view, previewMapView)
+    exportTilesButton.visibility = View.GONE
+
+    // // required for some Android devices running older OS (combats Z-ordering bug in Android API)
```
TADraeseke:
```suggestion
    // required for some Android devices running older OS (combats Z-ordering bug in Android API)
```

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,263 @@
...
+   */
+  private fun showMapPreview(result: TileCache) {
+    val newTiledLayer = ArcGISTiledLayer(result)
+    val previewMap = ArcGISMap(Basemap(newTiledLayer))
+
+    // set up the preview map view
+    previewMapView.apply {
+      map = previewMap
+      setViewpoint(mapView.getCurrentViewpoint(Viewpoint.Type.CENTER_AND_SCALE))
+    }
```
TADraeseke:
```suggestion

    // set up the preview map view
    previewMapView.apply {
      map = ArcGISMap(Basemap(ArcGISTiledLayer(result)))
      setViewpoint(mapView.getCurrentViewpoint(Viewpoint.Type.CENTER_AND_SCALE))
    }
```

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,263 @@
...
+    // control UI visibility
+    previewMapView.getChildAt(0).visibility = View.VISIBLE
+    show(closeButton, dimBackground, preview_text_view, previewMapView)
```
TADraeseke: preview_text_view should probably be in camel case like other view names

kotlin/export-tiles/src/main/java/com/example/exporttiles/MainActivity.kt
```diff
@@ -0,0 +1,263 @@
...
+   */
+  private fun createProgressDialog(exportTileCacheJob: ExportTileCacheJob): AlertDialog {
+    val builder = AlertDialog.Builder(this@MainActivity).apply {
+      setTitle("Exporting tiles...")
+      // provide a cancel button on the dialog
+      setNeutralButton("Cancel") { _, _ ->
+        exportTileCacheJob.cancel()
+      }
+      setView(R.layout.dialog_layout)
+    }
```
TADraeseke:
```suggestion
    val builder = AlertDialog.Builder(this@MainActivity).apply {
      setTitle("Exporting tiles...")
      // provide a cancel button on the dialog
      setNeutralButton("Cancel") { _, _ ->
        exportTileCacheJob.cancel()
      }
      setCancelable(false)
      setView(R.layout.dialog_layout)
    }
```

TADraeseke: Adding `setCancelable(false)` will make it so the user tapping outside the dialog or hitting back won't stop the download, but the "cancel" button will still work :)

Rachael-E:
@TADraeseke thanks very much for that! Changes made: could you take a final look before I merge please?

kotlin/export-tiles/README.md
```diff
@@ -0,0 +1,37 @@
+# Export tiles
+
+Download tiles to a local tile cache file stored on the device.
+
+![Export Tiles Sample](export-tiles.png)
```
TADraeseke: Was there a settled way to do this during the great README migration?

![Image of Export Tiles sample] or something like that?

kotlin/export-tiles/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+* ArcGISTiledLayer
+* ExportTileCacheJob
+* ExportTileCacheParameters
+* ExportTileCacheTask
+* TileCache
+
+## Additional information
+
+ArcGIS tiled layers do not support reprojection, query, select, identify, or editing. Visit the [ArcGiS Online Developer's portal](https://developers.arcgis.com/java/latest/guide/layer-types-described.htm#ESRI_SECTION1_BAA7882E7A7E4DE3AF4E9F137FA78279) to Learn more about the characteristics of ArcGIS tiled layers.
```
TADraeseke:
```suggestion
ArcGIS tiled layers do not support reprojection, query, select, identify, or editing. Visit the [ArcGIS Online Developer's portal](https://developers.arcgis.com/java/latest/guide/layer-types-described.htm#ESRI_SECTION1_BAA7882E7A7E4DE3AF4E9F137FA78279) to Learn more about the characteristics of ArcGIS tiled layers.
```

kotlin/export-tiles/README.md
Rachael-E: ah yes you're right, thanks for spotting that: should be just "image of export tiles", I've corrected it

PR #691, opened 2020-02-05
Hud10837/buffer kotlin

hud10837:
Rewrote the buffer sample in kotlin! Tried to keep it as similar as possible, but I'm not sure about code style or conciseness here. In theory the mapview setup at line 44 could be moved down and merged with the stuff in line 79, but I kept it better to match the java example and keep the map and graphics overlays separate.

Not sure if I'm using too much of the .apply syntax here, or whether there are smarter things to do here?

Thanks @TADraeseke for taking a look when you can


kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,148 @@
...
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_main)
+
+        // create a map with the Basemap
+        mapView.map = ArcGISMap(SpatialReferences.getWebMercator()).apply {
+            basemap = Basemap.createTopographic()
+        }
+
+        // create a fill symbol for geodesic buffer polygons
+        val geodesicOutlineSymbol = SimpleLineSymbol(SimpleLineSymbol.Style.SOLID, Color.BLACK, 2F)
```
TADraeseke: Could move the symbol creation lines close to the graphics overlays to which they're applied

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,148 @@
...
+
+                // get the point that was clicked and convert it to a point in the map
+                val screenPoint = android.graphics.Point(motionEvent.x.roundToInt(),
+                        motionEvent.y.roundToInt())
+                val mapPoint = mapView.screenToLocation(screenPoint)
+
+                // only draw a buffer if a value was entered
+                if (bufferInput.text.toString().isNotEmpty()) {
+                    // get the buffer distance (miles) entered in the text box.
+//                    val bufferInMiles = Double.valueOf(bufferInput.getText().toString())
```
TADraeseke: looks like this line can be removed

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,148 @@
...
+        clearButton.setOnClickListener {
+            planarGraphicsOverlay.graphics.clear()
+            geodesicGraphicsOverlay.graphics.clear()
+            tapLocationsOverlay.graphics.clear()
+        }
+    }
+
+    override fun onPause() {
+        super.onPause()
+        mapView.pause()
```
TADraeseke: We just try to order the lifecycle calls like in here: https://devtopia.esri.com/runtime/common-samples/wiki/Android-Samples-Repo%3A-A-Contribution-Guide#handle-life-cycles

PR #693, opened 2020-02-06
Hud10837/buffer kotlin

hud10837:
@TADraeseke made the changes you requested and fixed all the wild gradle issues. Should be all good unless you notice anything new!


kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,160 @@
...
+    clearButton.setOnClickListener {
+      planarGraphicsOverlay.graphics.clear()
+      geodesicGraphicsOverlay.graphics.clear()
+      tapLocationsOverlay.graphics.clear()
+    }
+  }
+
+  override fun onResume() {
+    super.onResume()
+    mapView.resume()
```
TADraeseke: order in lifecycle aware way https://devtopia.esri.com/runtime/common-samples/wiki/Android-Samples-Repo%3A-A-Contribution-Guide#handle-life-cycles

kotlin/build.gradle
```diff
@@ -10,7 +12,7 @@ buildscript {
     dependencies {
         classpath "com.android.tools.build:gradle:$gradleVersion"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-    }
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version" }
```
TADraeseke: this got auto-updated I think and can be removed

kotlin/build.gradle
```diff
@@ -1,16 +1,17 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 apply from: '../version.gradle'
 
-buildscript {
+buildscript { 
+    ext.kotlin_version = '1.3.50'
```
TADraeseke: These can be removed--I think Android Studio threw these in automatically

kotlin/buffer/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Additional information
+
+Buffers can be generated as either `planar` (flat - coordinate space of the map's spatial reference) or `geodesic` (technique that considers the curved shape of the Earth's surface, which is generally a more accurate representation). In general, distortion in the map increases as you move away from the standard parallels of the spatial reference's projection. This map is in Web Mercator so areas near the equator are the most accurate. As you move the buffer location north or south from that line, you'll see a greater difference in the polygon size and shape. Planar operations are generally faster, but performance improvement may only be noticeable for large operations (buffering a great number or complex geometry).
+
+For more information about using buffer analysis, see [How buffer analysis works](https://pro.arcgis.com/en/pro-app/tool-reference/analysis/how-buffer-analysis-works.htm) in the ArcGIS Pro documentation.  
+
+## Tags
+
+Analysis, Planar, GeometryEngine, BufferGeodetic, Geodesic, Euclidean
```
Rachael-E: I think this readme might have been one of the earlier ones to be refactored in the common-samples repo (there's some history to this before you joined: the process evolved over time!). Tags should be mostly lower case and not include API references if already mentioned in the Relevant API section. I'd suggest the following changes, and we'd also have to make this change in the original design in the common-sample repo to keep it in line with the final format the team agreed on: makes sense to do this as we go and notice these things! 
 
```suggestion
analysis, planar, geodesic, euclidean
```

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+import kotlinx.android.synthetic.main.activity_main.*
+import kotlin.math.roundToInt
+
+class MainActivity : AppCompatActivity() {
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_main)
+
+    // create a map with the Basemap
```
Rachael-E:
```suggestion
    // create a map with a topographic basemap
```

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+    // create a marker symbol for tap locations
+    val tapSymbol = SimpleMarkerSymbol(SimpleMarkerSymbol.Style.CROSS, Color.WHITE, 14F)
+
+    // create a graphics overlay to display tap locations for buffers and set its renderer
+    val tapLocationsOverlay = GraphicsOverlay().apply {
+      renderer = SimpleRenderer(tapSymbol)
+    }
+
+    // add overlays to the mapView
+    mapView.graphicsOverlays.apply {
```
Rachael-E: I think we should use this instead of the apply, a bit neater:

`mapView.graphicsOverlays.addAll(listOf(geodesicGraphicsOverlay, planarGraphicsOverlay, tapLocationsOverlay))`

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+      add(geodesicGraphicsOverlay)
+      add(planarGraphicsOverlay)
+      add(tapLocationsOverlay)
+    }
+
+    // create a buffer around the clicked location
+
```
Rachael-E: can remove white space here

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+          Toast.makeText(
+            this@MainActivity,
+            "Please enter a buffer distance first.",
+            Toast.LENGTH_LONG
+          ).show()
+        }
+        return true
+      }
+    }
+
```
Rachael-E: let's add a comment here too, something simple like
`clear the graphics from the graphics overlays`

kotlin/buffer/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="colorPrimary">#3F51B5</color>
```
Rachael-E: let's use the standardized colors as specified in the contribution guide

```
    <color name="colorPrimary">#2196F3</color>
    <color name="colorPrimaryDark">#1976D2</color>
    <color name="colorAccent">#4CAF50</color>
```

kotlin/buffer/README.md
hud10837: I'll keep that in mind for the future!

kotlin/buffer/src/main/java/com/esri/arcgisruntime/sample/buffer/MainActivity.kt
hud10837: this is definitely neater, I'm trying to avoid using apply too much now

hud10837:
That's those changes made!

Rachael-E:
thanks @hud10837! I've just noticed that the metadata.json file is missing from this PR: please could you add one in. These files exist for easier searching of samples on our webpage and are important to include.  

You should mostly be able to just copy the one over from the Android Java repo (https://github.com/Esri/arcgis-runtime-samples-android/blob/master/java/buffer/README.metadata.json) though you'll have to check and update it based on the newer readme you have in this PR. Keywords are equivalent to tags (and should be the same case as in the readme), and you'll have to update the description (to be the same as the summary in the updated readme). The "snippets section" will also need to be updated from MainActivity.java to MainActivity.kt. Any questions just let me know!

hud10837:
@Rachael-E thanks, I think that's correct now!

PR #696, opened 2020-02-10
Sor10874/459 viewshed location kotlin

sorenoid:
kotlin nicely cuts quite a bit of ui initialization out of this sample. no major changes from java. 

I made the choice to stick with Ints for all the directional variable (pitch, angle etc) since even though they are consumed by the viewshed as doubles, they are produced by the UI controls as Ints. So there is no loss and there are fewer casts this way. 

It was fun to play around with this one!


kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,354 @@
...
+    ).apply {
+      setFrustumOutlineVisible(true)
+    }
+
+    // add a camera and set it to orbit the location point of the frustum
+    val camera = Camera(initLocation, 20000000.0, 0.0, 55.0, 0.0)
+    val orbitCamera = OrbitLocationCameraController(initLocation, 5000.0)
+    sceneView.apply {
+      this.scene = scene
+      this.cameraController = orbitCamera
```
TADraeseke: Possibly easier to just inline `Camera(initLocation, 20000000.0, 0.0, 55.0, 0.0)`

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,354 @@
...
+      this.setViewpointCamera(camera)
+    }
+
+    // create an analysis overlay to add the viewshed to the scene view
+    val analysisOverlay = AnalysisOverlay()
+    analysisOverlay.analyses.add(viewShed)
```
TADraeseke: I'd find it more readable using scoping function myself:
```
    val analysisOverlay = AnalysisOverlay().apply { 
      analyses.add(viewShed)
    }
    sceneView.analysisOverlays.add(analysisOverlay)
```

But others would disagree--so whichever you prefer :)

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,354 @@
...
+        )
+        val locationPointFuture = sceneView.screenToLocationAsync(screenPoint)
+        locationPointFuture.addDoneListener {
+          try {
+            val locationPoint = locationPointFuture.get()
+
+            // add 50 meters to location point and set to viewshed
+            viewShed.location = Point(locationPoint.x, locationPoint.y, locationPoint.z + 50)
+          } catch (e: InterruptedException) {
+            val error = "Error converting screen point to location point: " + e.message
```
TADraeseke: I likes the .also pattern you used on other samples...could be good here again!

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
sorenoid: did a bit more cleanup here by renaming scene to something more meaningful (buildingsScene) which also allowed me to drop the this. from the calls in the scope function

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
sorenoid: i did make this decision based on the number of lines of code required basis, but i think you are right about the readability.

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
sorenoid: I changed it up a bit and added a method to log and toast, like the contributor's guide says, and used let{} so i could easily handle null strings.

kotlin/viewshed-location/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+![Image of viewshed location](viewshed-location.png)
+
+## Use case
+
+A 3D viewshed analysis is a type of visual analysis you can perform on a scene. The viewshed shows what can be seen from a given location. The output is an overlay with two different colors - one representing the visible areas (green) and the other representing the obstructed areas (red). Viewshed analysis is a form of "exploratory analysis", which means the results are calculated on the current scale of the data, and the results are generated very quickly. If more "conclusive" results are required, consider using a `GeoprocessingTask` to perform a viewshed instead.
+
+## How to use the sample
+
+Use the sliders to change the properties (heading, pitch, etc.), of the viewshed and see them updated in real time. To move the viewshed, double touch and drag your finger across the screen. Lift your finger to stop moving the viewshed.
```
Rachael-E: looking at the common design png (and after comparing with the Java SE sample), it looks as though we're missing the option to turn the visibility of both the viewshed and the frustum on and off. Apologies if I've missed a reason why that wasn't included: I know that our mobile apps are more restricted for UI than desktop (e.g. it looks like Java SE went one step further than the design and added in additional controls to change the colour of the viewshed!)

kotlin/viewshed-location/README.md
```diff
@@ -0,0 +1,34 @@
...
+* LocationViewshed
+* Viewshed
+
+## About the data
+
+The scene in this sample shows a [buildings layer](https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0) in Brest, France.
+
+## Tags
+
+visibility analysis, viewshed, LocationViewshed, AnalysisOverlay, Scene, 3D, frustum
```
Rachael-E: looks like this is another example from the common-sample readme which contains references to the API in the tags: let's remove LocationViewshed and AnalysisOverlay since they are already in the Relevant API section. I think it's ok to leave Scene in here though (it's not the focus of the sample, so it doesn't get put in the Relevant API section, but folk who search for "Scene" could still find this).

kotlin/viewshed-location/README.metadata.json
```diff
@@ -0,0 +1,29 @@
+{
+    "category": "Analysis",
+    "description": "A viewshed shows the visible and obstructed areas from an observer's vantage point. This sample demonstrates all of the configurable properties of a viewshed.",
```
Rachael-E: This description should match the summary from the readme
```suggestion
    "description": "Perform a viewshed analysis from a defined vantage point.",
```

kotlin/viewshed-location/README.metadata.json
```diff
@@ -0,0 +1,29 @@
...
+    "ignore": false,
+    "images": [
+        "viewshed-location.png"
+    ],
+    "keywords": [
+        "visibility analysis",
+        "viewshed",
+        "LocationViewshed",
```
Rachael-E: can remove LocationViewshed and AnalysisOverlay as mentioned above

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,355 @@
...
+import kotlinx.android.synthetic.main.viewshed_seek_bars.*
+
+
+class MainActivity : AppCompatActivity() {
+
+  companion object {
+    private val TAG: String = MainActivity::class.java.simpleName
+  }
+
+  private val initHeading = 0
```
Rachael-E: a wee comment here would be good, something like:
`// initialize location viewshed parameters`

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,355 @@
...
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_main)
+
+    // create a surface for elevation data
+    val surface = Surface().apply {
+      elevationSources.add(ArcGISTiledElevationSource(getString(R.string.elevation_service)))
+    }
+
+    // create a layer of buildings
+    val buildingsLayer = ArcGISSceneLayer(getString(R.string.buildings_layer))
```
Rachael-E: minor change here, I think it would be good to rename this to `buildingsSceneLayer`, just to make it clear to the user what type of layer this is throughout the code when they see the variable. We have different types of layers throughout our API, and keeping the variable name as clear as we can is helpful for our customers.

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,355 @@
...
+
+    // create a scene and add imagery basemap, elevation surface, and buildings layer to it
+    val buildingsScene = ArcGISScene().apply {
+      basemap = Basemap.createImagery()
+      baseSurface = surface
+      operationalLayers.add(buildingsLayer)
+    }
+
+    val initLocation = Point(-4.50, 48.4, 1000.0)
+    Viewshed.setFrustumOutlineColor(Color.BLUE)
```
Rachael-E: this is probably personal preference, but I think this line would read better immediately after the viewShed set up

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,355 @@
...
+      initHorizontalAngle.toDouble(),
+      initVerticalAngle.toDouble(),
+      initMinDistance.toDouble(),
+      initMaxDistance.toDouble()
+    ).apply {
+      setFrustumOutlineVisible(true)
+    }
+
+    sceneView.apply {
+      //add the buildings scene to the sceneView
```
Rachael-E:
```suggestion
      // add the buildings scene to the sceneView
```

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,355 @@
...
+      initPitch.toDouble(),
+      initHorizontalAngle.toDouble(),
+      initVerticalAngle.toDouble(),
+      initMinDistance.toDouble(),
+      initMaxDistance.toDouble()
+    ).apply {
+      setFrustumOutlineVisible(true)
+    }
+
+    sceneView.apply {
```
Rachael-E: where possible, we try to call the scene/sceneView set up as one of the first few steps in our samples: think of it like the foundation set up steps before leading on to the main focus of what the sample is for. therefore here, I'd recommend moving the `sceneView.apply` block just before the `viewShed` initialization block.

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
```diff
@@ -0,0 +1,355 @@
...
+      //add the buildings scene to the sceneView
+      scene = buildingsScene
+      // add a camera and set it to orbit the location point of the frustum
```
Rachael-E: I noticed that if you double tap and drag the frustum away, the camera doesn't orbit the location point of the frustum. therefore we might want to alter this comment slightly to 
```suggestion
      // add a camera and set it to orbit the starting location point of the frustum
```

kotlin/viewshed-location/README.md
Rachael-E: just noticed it was missing from the original Android java sample too: I think for parity with other SDKs, we should include those check boxes for visibility and frustum if it's not too overwhelming on the UI. e.g. from Java SE:

```
visibilityToggle.selectedProperty().addListener(e -> viewshed.setVisible(visibilityToggle.isSelected()));
```

```
frustumToggle.selectedProperty().addListener(e -> viewshed.setFrustumOutlineVisible(frustumToggle.isSelected()))
```

kotlin/viewshed-location/src/main/java/com/esri/arcgisruntime/sample/viewshedlocation/MainActivity.kt
sorenoid: i apologize for missing these whitespace things.

PR #697, opened 2020-02-11
Sor10874/93 identify layers kotlin

sorenoid:
this one is fairly straight forward, though the code performs a depth first search through all layers' sublayers for elements. It is fairly hard to find a case where a layer has any sublayers, the tiny city of Annapolis, Maryland (near Washington D.C. on the east coast) seems to invoke the full search behavior consistently.


kotlin/identify-layers/src/main/java/com/esri/arcgisruntime/sample/identifylayers/MainActivity.kt
```diff
@@ -0,0 +1,218 @@
...
+  private fun identifyResult(screenPoint: android.graphics.Point) {
+
+    val identifyLayerResultsFuture = mapView
+      .identifyLayersAsync(screenPoint, 12.0, false, 10)
+
+    identifyLayerResultsFuture.addDoneListener {
+      try {
+        val identifyLayerResults = identifyLayerResultsFuture.get()
+        handleIdentifyResults(identifyLayerResults)
+      } catch (e: InterruptedException) {
```
TADraeseke: Can just use 'Exception' if returning the same message

kotlin/identify-layers/README.md
```diff
@@ -0,0 +1,34 @@
...
+2. For each `IdentifyLayerResult` in the results, features are counted. 
+    * Note: there is one identify result per layer with matching features; if the feature count is 0, that means a sublayer contains the matching features.
+
+## Relevant API
+
+* `IdentifyLayerResult`
+* `MapView.IdentifyLayersAsync`
+* `IdentifyLayerResult.LayerContent.Name`
+* `IdentifyLayerResult.SublayerResults`
+
```
Rachael-E: let's move the MapView call to the end for alphabetization (this will have probably been caught in Zack's updates to the common Readmes too)

kotlin/identify-layers/README.md
```diff
@@ -0,0 +1,34 @@
+# Identify layers
+
+Identify features in all layers in a map. 
+
+![Image of identify layers](identify-layers.png)
+
+## Use case
+
+IdentifyLayers allows users to tap on a map, returning features at that location across multiple layers. Because some layer types have sublayers, the sample recursively counts results for sublayers within each layer.
```
Rachael-E: minor readme improvement, we generally try to avoid calling API names in the use case: this one must have slipped through the common sample readme net!
```suggestion
Identify layers allows users to tap on a map, returning features at that location across multiple layers. Because some layer types have sublayers, the sample recursively counts results for sublayers within each layer.
```

kotlin/identify-layers/README.metadata.json
```diff
@@ -0,0 +1,25 @@
...
+  "keywords": [
+    "identify",
+    "recursion",
+    "recursive",
+    "sublayers"
+  ],
+  "redirect_from": "/android/latest/sample-code/identify-layers.htm",
+  "relevant_apis": [
+    "IdentifyLayerResult",
+    "MapView.IdentifyLayersAsync",
```
Rachael-E: update to be alphabetical too

PR #701, opened 2020-02-14
Sor10874/192 oauth2 kotlin

sorenoid:
ok this one threw me for a loop because it wasn't behaving very well but it turned out to be a developer setting I had turned on. It might still be an issue for arcgis-android, but that shouldn't block this sample! If you do have the dev setting "don't keep activities" turned on, please turn it off before running this app (same or the java and released version).

one thing to note is that I treated OAuth as a proper noun in the title and left it capitalized on advice of counsel (Mike).


kotlin/authenticate-with-oauth/src/main/java/com/esri/arcgisruntime/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -0,0 +1,107 @@
...
+
+  override fun onResume() {
+    super.onResume()
+    mapView.resume()
+  }
+
+  override fun onPause() {
+    // normally, you won't want to clear credentials once a device has been verified. These calls are made to keep this
+    // sample from interfering with other authentication samples
+    AuthenticationManager.CredentialCache.clear();
```
TADraeseke: can lose the ; on this and the next line

It was my idea to put this in OnPause in the original Java sample, but I wonder if it might more usefully be placed at the start of onCreate--This would reset credentials each time the app is run without it being reset if someone just loses focus of the app

kotlin/authenticate-with-oauth/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Additional information
+
+The workflow presented in this sample works for all SAML based enterprise (IWA, PKI, Okta, etc.) & social (facebook, google, etc.) identity providers for ArcGIS Online or Portal. For more information, see the topic [Set up enterprise logins].(https://doc.arcgis.com/en/arcgis-online/administer/enterprise-logins.htm).
+
+For additional information on using OAuth in your app, see the topic [Authenticate with the API](https://developers.arcgis.com/documentation/core-concepts/security-and-authentication/mobile-and-native-user-logins/) in *Mobile and Native Named User Login*.
+
+## Tags
+
+cloud & portal, authentication, security, oauth, credential
```
Rachael-E: these are the updated tags based on comparison with our big readme effort: main thing is alphabetisation and the separation of cloud and portal. 

authentication, cloud, credential, OAuth, portal, security

kotlin/authenticate-with-oauth/README.md
```diff
@@ -0,0 +1,37 @@
...
+## How it works
+
+1. Set the `AuthenticationManager`'s `AuthenticationChallengeHandler` to a `DefaultAuthenticationChallengeHandler`.
+1. Create an `OAuthConfiguration` specifying the portal URL, client ID, and redirect URL.
+1. Add the OAuth configuration to the authentication manager.
+1. Load a map with premium content requiring authentication to automatically invoke the default authentication handler.
+
+## Relevant API
+
+ * AuthenticationManager
```
Rachael-E: I think I'm on the most up to date version of this branch, but I'm still seeing these bullet points as stepped and not inline as mentioned by Trevor

kotlin/authenticate-with-oauth/README.metadata.json
```diff
@@ -0,0 +1,26 @@
+{
+  "category": "Cloud and Portal",
+  "description": "Authenticate with ArcGIS Online (or your own portal) using OAuth2 to access secured resources (such as private web maps or layers). Accessing secured items requires logging in to the portal that hosts them (an ArcGIS Online account, for example).",
+  "ignore": false,
+  "images": [
+    "authenticate-with-oauth.png"
+  ],
+  "keywords": [
```
Rachael-E: I guess since we're here might as well update the keywords based on tag changes above?

sorenoid:
@Rachael-E I've addressed your suggestions, ok to merge?

PR #702, opened 2020-02-17
Hud10837/buffer kotlin units

hud10837:
This is just a tiny change to the visuals on this sample; added "miles" after the input field as we discussed in the review.

### DIFF
--- a/kotlin/buffer/src/main/res/layout/activity_main.xml
+++ b/kotlin/buffer/src/main/res/layout/activity_main.xml
@@ -25,8 +25,18 @@
         android:hint="@string/Enter_miles_string"
         android:inputType="number"
         android:text="@string/default_buffer_distance"
-        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintEnd_toStartOf="@id/unitsDisplay"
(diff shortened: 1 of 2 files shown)


kotlin/buffer/src/main/res/layout/activity_main.xml
```diff
@@ -25,8 +25,18 @@
...
         android:inputType="number"
         android:text="@string/default_buffer_distance"
-        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintEnd_toStartOf="@id/unitsDisplay"
         app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <TextView
+        android:id="@+id/unitsDisplay"
+        android:layout_width="="
```
Rachael-E: I'm getting an `Android resource linking failed` error with this line: the following seems to be ok though?
```suggestion
        android:layout_width="wrap_content"
```

hud10837: That's really strange, I've never seen "=" as a width. How did that happen?

PR #707, opened 2020-02-19
Add point to add with renderer

hud10837:
Hi all!
- We already discussed the directory/readme names; I used the common readme titles for the new readmes.

- The screenshot names were pretty inconsistent within the common designs.

- I kept the `How it works` section from the new readmes, because I thought they were better, but let me know if you disagree.
Thanks!


java/add-point-scene-layer/README.md
```diff
@@ -1,8 +1,8 @@
-# Add point scene layer
+# Add a point scene layer
 
 View a point scene layer from a scene service.
 
-![Add point scene layer App](add-point-scene-layer.png)
+![Image of a point scene layer](AddAPointSceneLayer.png)
```
Rachael-E: let's not change the .png file names: if we do that then we'd have to update the png file name locally too and that adds extra overload.

kotlin/add-features-feature-service/README.md
```diff
@@ -1,33 +1,34 @@
-# Add features (Feature Service)
+# Add features
```
Rachael-E: I think this is an example of the scenario Jon mentioned earlier and then what Lucas clarified: let's leave this title as it was "Add features (Feature Service)" and we can catch any changes after the next Readme Title effort.

kotlin/add-features-feature-service/README.md
```diff
@@ -1,33 +1,34 @@
...
 
 ## Use case
 
-A `FeatureLayer` created using this `ServiceFeatureTable` that is applied to the `ArcGISMap` will display any new Features automatically.
+An end-user performing a survey may want to add features to the map during the course of their work.
 
 ## How to use the sample
 
-Tap on the map to add a feature.
+Click on a location on the map to add a feature at that location.
```
Rachael-E: this is going to be one I know I'll be guilty of too: mentioning click vs tap! ;) 
```suggestion
Tap on a location on the map to add a feature at that location.
```

kotlin/integrated-mesh-layer/README.md
```diff
@@ -2,26 +2,36 @@
...
+3. Add the layer to the scene's operational layers.
 
 ## Relevant API
 
-- IntegratedMeshLayer
+* IntegratedMeshLayer
+
+## About the data
+
+This data is derived from  point cloud data from a lidar survey. Lidar data is often of a high enough resolution to capture 3D features as small as walls and trees, but also the irregular surfaces of geological features like cliffs. An integrated mesh made from a point cloud of lidar data will include realistic textures and elevation information.
```
Rachael-E: super nit picky alert: extra space can be removed
```suggestion
This data is derived from point cloud data from a lidar survey. Lidar data is often of a high enough resolution to capture 3D features as small as walls and trees, but also the irregular surfaces of geological features like cliffs. An integrated mesh made from a point cloud of lidar data will include realistic textures and elevation information.
```

hud10837:
Thanks guys, that's those changes made!

PR #709, opened 2020-02-19
Sor10874/readmes/1 4

sorenoid:
this PR includes 

feature-layer-geodatabase
feature-layer-geopackage
feature-layer-query (java and kotlin)
feature-layer-rendering-mode-map

some small changes to the geodatabase text because it doesn't load a vector tile basemap as the common sample says, and changes to the feature-layer-query to change click to tap (and my keybaord says "GO", not "Search" so I changed that text in the README as well).

### DIFF
--- a/java/feature-layer-geopackage/README.md
+++ b/java/feature-layer-geopackage/README.md
@@ -1,19 +1,31 @@
-# Feature Layer (GeoPackage)
-Open a GeoPackage and show a GeoPackage feature table in a feature layer.
+# Feature layer (GeoPackage)
 
-![Feature layer GeoPackage App](feature-layer-geopackage.png)
+Display features from a local GeoPackage.
+
+![Image of feature layer geopackage](feature-layer-geopackage.png)
+
+## Use case
+
+A GeoPackage is an OGC standard, making it useful when your project requires an open source data format or when other, non-ArcGIS systems may be creating the data. Accessing data from a local GeoPackage is useful when working in an environment that has an inconsistent internet connection or that does not have an internet connection at all. For example, a department of transportation field worker might source map data from a GeoPackage when conducting signage inspections in rural areas with poor network coverage.
 
 ## How to use the sample
-Provision the device with local data as per the 'Offline Data' section below. Run the app and accept read permissions.
+
+Pan and zoom around the map. View the data loaded from the geopackage.
 
 ## How it works
-1. Create a `GeoPackage` from a local path and load it with `loadAsync()`.
-1. On successful load, get a `FeatureTable` from the `GeoPackage` using `getGeoPackageFeatureTables()`.
+
+1. Create a `GeoPackage` passing the URI string into the constructor.
+2. Load the `GeoPackage` with `GeoPackage.loadAsync`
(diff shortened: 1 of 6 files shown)


java/feature-layer-geopackage/README.md
```diff
@@ -1,19 +1,31 @@
...
 
 ## How it works
-1. Create a `GeoPackage` from a local path and load it with `loadAsync()`.
-1. On successful load, get a `FeatureTable` from the `GeoPackage` using `getGeoPackageFeatureTables()`.
+
+1. Create a `GeoPackage` passing the URI string into the constructor.
+2. Load the `GeoPackage` with `GeoPackage.loadAsync`
+3. When it's done loading, get the `GeoPackageFeatureTable` objects from the geopackage with `geoPackage.getGeoPackageFeatureTables()`
+4. Create a `FeatureLayer(featureTable)` for each feature table and add it to the map as an operational layer. Add each to 
+   the map as an operational layer with `map.getOperationalLayers().add(featureLayer)`
```
hud10837: Is this a rogue newline?

java/feature-layer-query/README.md
```diff
@@ -1,18 +1,35 @@
...
-* FeatureTable
-* GeoPackage
+* FeatureQueryResult
+* QueryParameters
+* ServiceFeatureTable
+
+## Tags
 
-#### Tags
 Search and Query
```
hud10837: This is definitely what the common design says, but it seems like the tags should be different. @Rachael-E do you have any idea?

kotlin/feature-layer-query/README.md
```diff
@@ -2,15 +2,15 @@
...
 ## Use case
 
 Query expressions can be used in ArcGIS to select a subset of features from a feature table. This is most useful in large or complicated data sets. A possible use case might be on a feature table marking the location of street furniture through a city. A user may wish to query by a TYPE column to return "benches". In this sample, we query a U.S. state by STATE_NAME from a feature table containing all U.S. states.
 
 ## How to use the sample
 
-Input the name of a U.S. state into the text field. When you click "search", a query is performed and the matching features are highlighted or an error is returned.
+Input the name of a U.S. state into the text field. When you tap "GO", a query is performed and the matching features are highlighted or an error is returned.
 
 ## How it works
```
Rachael-E: For step 2 for Kotlin, we should replace `setWhereClause` with `whereClause` so that we are following Kotlin syntax

java/feature-layer-query/README.md
Rachael-E: this is ok to leave as is, it's a standard search term from what i understand from previous discussions.

Rachael-E:
@sorenoid great attention to detail spotting we don't use the vector tile basemap! I checked and it's the same for the Java SDK sample, some SDKs must vary slightly in how they're showing that basemap. Since the basemap isn't the main point of the sample that's ok. 
> some small changes to the geodatabase text because it doesn't load a vector tile basemap as the common sample says

java/feature-layer-geopackage/README.md
sorenoid: I saw that and wondered why it was there, everything else just goes on forever. I'll remove it.

PR #718, opened 2020-02-21
Readmes rach kotlin1

Rachael-E:
I got a little carried away and accidentally changed 7 samples. Oops! The next one will just be 3 to make up for it >.< ...decided to do all the kotlin ones first. 

Only thing to really note is that Scene Layer has a minor tweak in that I've removed reference to the adding in of a surface since we don't do that in our samples. @sorenoid could you take a first look please? thanks!

### DIFF
--- a/kotlin/raster-function-service/README.md
+++ b/kotlin/raster-function-service/README.md
@@ -1,54 +1,43 @@
-# Raster Function Service
+# Raster function service
 
-Raster functions are operations performed on a `Raster`to apply on-the-fly processing such as Hillshade.  This sample demonstrates how to create an `ImageServiceRaster` and apply a `RasterFunction` to it. The `RasterFunction` applied in this case is Hillshade.
+Load a raster from a service, then apply a function to it.
 
 ![Raster Function App](raster-function-service.png)
 
-## Features
+## Use case
+
+Raster functions allow processing operations that can be applied to one or more rasters on the fly. Functions can be applied to rasters that come from a service. A land survey agency may apply hillshade and aspect functions to rasters with elevation data in order to better determine the topography of a landscape and to make further planning decisions.
+
+## How to use the sample
+
+The raster function is applied automatically when the sample starts and the result is displayed.
+
+## How it works
+
+1. Create the `ImageServiceRaster` referring to the image server URL.
+2. Create the `RasterFunction` from a JSON string.
+3. Get the name of the raster argument to the function with `rasterFunction.arguments.rasterNames[0]`
+4. Set the raster argument with `rasterFunction.setRaster(rasterName, imageServiceRaster)`.
+5. Create a new `Raster` referring to the raster function.
+6. Create a `RasterLayer` to visualize the computed raster.
+7. Display the raster.
+
+## Relevant API
 
 * ImageServiceRaster
+* Raster
 * RasterFunction
+* RasterFunctionArguments
 * RasterLayer
 
-## Developer Pattern
+## About the data
 
-To create a `Raster` from a `RasterFunction`:
+The sample applies a hillshade function to a raster produced from the National Land Cover Database, [NLCDLandCover2001](https://sampleserver6.arcgisonline.com/arcgis/rest/services/NLCDLandCover2001/ImageServer). You can learn more about the [hillshade function](http://desktop.arcgis.com/en/arcmap/latest/manage-data/raster-and-images/hillshade-function.htm) in the *ArcMap* documentation.
 
-* Create a `RasterFunction` from a json string resource.
-* Create a `RasterFunctionArguments` from the `RasterFunction`.
-* Set a `Raster` with a name parameter in `RasterFunctionArguments`
-* Create a new `Raster` from the `RasterFunction`.
-* Add it as an operational layer with `mapView.map.operationalLayers.add(hillshadeLayer)`.
+## Additional information
 
-```kotlin
-fun applyRasterFunction(raster: Raster){
-    // create raster function from json string
-    val rasterFunction = RasterFunction.fromJson(resources.getString(R.string.hillshade_simplified))
-    // get parameter name value pairs used by hillshade
-    val rasterFunctionArguments = rasterFunction.arguments
-    // get list of raster names associated with raster function
-    val rasterName = rasterFunctionArguments.rasterNames
-    // set raster to the raster name
-    rasterFunctionArguments.setRaster(rasterName[0], raster)
-    // create raster as raster layer
-    val raster = Raster(rasterFunction)
-    val hillshadeLayer = RasterLayer(raster)
-    // add hillshade raster
-    mapView.map.operationalLayers.add(hillshadeLayer)
-}
-```
+The raster function computation happens locally on the client device. 
 
-The sample also zooms to the extent of the `ImageServiceRaster`.  Currently we do not support zooming a `RasterLayer` out beyond 4 times it's published level of detail. The sample uses `MapView.setViewpointCenterAsync()` method to ensure the image shows when the app starts. You can see the effect of the image service not showing when you zoom out to the full extent of the image and beyond. 
+## Tags
 
-```kotlin
-// zoom to the extent of the raster service
-imageRasterLayer.addDoneLoadingListener {
-    if(imageRasterLayer.loadStatus == LoadStatus.LOADED){
-        // zoom to extent of raster
-        val centerPnt = imageServiceRaster.serviceInfo.fullExtent.center
-        mapView.setViewpointCenterAsync(centerPnt, 55000000.0)
-        // update raster with simplified hillshade
-        applyRasterFunction(imageServiceRaster)
-    }
-}
-```
+function, layer, raster, raster function, service
--- a/kotlin/raster-function-service/README.md
+++ b/kotlin/raster-function-service/README.md
@@ -16,7 +16,7 @@ The raster function is applied automatically when the sample starts and the resu
 
 1. Create the `ImageServiceRaster` referring to the image server URL.
 2. Create the `RasterFunction` from a JSON string.
-3. Get the name of the raster argument to the function with `rasterFunction.arguments.rasterNames[0]`
+3. Get the name of the raster argument to the function with `rasterFunction.arguments.rasterNames[0]`.
 4. Set the raster argument with `rasterFunction.setRaster(rasterName, imageServiceRaster)`.
 5. Create a new `Raster` referring to the raster function.
 6. Create a `RasterLayer` to visualize the computed raster.
--- a/kotlin/raster-function-service/README.md
+++ b/kotlin/raster-function-service/README.md
@@ -2,7 +2,7 @@
 
 Load a raster from a service, then apply a function to it.
 
-![Image of apply raster function to raster from service](raster-function-service.png)
+![Image of raster function service](raster-function-service.png)
 
 ## Use case
 
--- a/kotlin/raster-function-service/README.md
+++ b/kotlin/raster-function-service/README.md
@@ -10,13 +10,14 @@ Raster functions allow processing operations that can be applied to one or more 
 
 ## How to use the sample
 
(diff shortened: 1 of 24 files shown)


Rachael-E:
thanks @sorenoid ! missing period added. @hud10837 could you take a second look please?

kotlin/raster-function-service/README.md
```diff
@@ -1,54 +1,43 @@
-# Raster Function Service
+# Raster function service
 
-Raster functions are operations performed on a `Raster`to apply on-the-fly processing such as Hillshade.  This sample demonstrates how to create an `ImageServiceRaster` and apply a `RasterFunction` to it. The `RasterFunction` applied in this case is Hillshade.
+Load a raster from a service, then apply a function to it.
 
-![Raster Function App](raster-function-service.png)
+![Image of apply raster function to raster from service](raster-function-service.png)
```
hud10837: Should this be `Image of raster function service` to match the sample name? 
```suggestion
![Image of raster function service](raster-function-service.png)
```

kotlin/raster-function-service/README.md
```diff
@@ -1,54 +1,43 @@
...
 
-## Features
+## Use case
+
+Raster functions allow processing operations that can be applied to one or more rasters on the fly. Functions can be applied to rasters that come from a service. A land survey agency may apply hillshade and aspect functions to rasters with elevation data in order to better determine the topography of a landscape and to make further planning decisions.
+
+## How to use the sample
+
+The raster function is applied automatically when the sample starts and the result is displayed.
```
hud10837: It looks like the function for this sample is tied to a `Apply Layer` button?

kotlin/raster-function-service/README.md
```diff
@@ -1,54 +1,43 @@
...
+
+## How it works
+
+1. Create the `ImageServiceRaster` referring to the image server URL.
+2. Create the `RasterFunction` from a JSON string.
+3. Get the name of the raster argument to the function with `rasterFunction.arguments.rasterNames[0]`.
+4. Set the raster argument with `rasterFunction.setRaster(rasterName, imageServiceRaster)`.
```
hud10837: Looking at the code it seems like this is a bit different from common; lines 80-90

```
// create raster function from json string
    RasterFunction rasterFuntionFromJson = RasterFunction.fromJson(getString(R.string.hillshade_simplified));
    // get parameter name value pairs used by hillside
    RasterFunctionArguments rasterFunctionArguments = rasterFuntionFromJson.getArguments();
    // get a list of raster names associated with the raster function
    List<String> rasterNames = rasterFunctionArguments.getRasterNames();
    rasterFunctionArguments.setRaster(rasterNames.get(0), raster);
    // create raster as raster layer
    raster = new Raster(rasterFuntionFromJson);
    RasterLayer hillshadeLayer = new RasterLayer(raster);
    mMapView.getMap().getOperationalLayers().add(hillshadeLayer);
```

I would suggest:

```
3. Get the arguments of the raster function with RasterFunction.getArguments()
4. Get the names of the rasters with RasterFunctionArguments.getRasterNames()
5. Set the raster argument with `RasterFunctionArguments.setRaster(rasterName, imageServiceRaster)
```

But maybe there's a way to shorten that back down to two steps.

kotlin/raster-rendering-rule/README.md
```diff
@@ -1,34 +1,36 @@
...
+# Raster rendering rule
+
+Display a raster on a map and apply different rendering rules to that raster.
+
+![Image of raster rendering rule](raster-rendering-rule.png)
+
+## Use case
+
+Raster images whose individual pixels represent elevation values can be rendered in a number of different ways, including representation of slope, aspect, hillshade, and shaded relief. Applying these different rendering rules to the same raster allows for a powerful visual analysis of the data. For example, a geologist could interrogate the raster image to map subtle geological features on a landscape, which may become apparent only through comparing the raster when rendered using several different rules.
+
```
hud10837: Should there be a "How to use this sample" section here?

kotlin/raster-rendering-rule/README.md
```diff
@@ -1,34 +1,36 @@
...
+## How it works
+
+1.  Create an `ImageServiceRaster` using a URL to an online image service.
+2.  After loading the raster, use `imageServiceRaster.serviceInfo().renderingRuleInfos()` to get a list of `RenderingRuleInfo` supported by the service.
```
hud10837: Should these be class names rather than instance names?
```suggestion
2.  After loading the raster, use `ImageServiceRaster.serviceInfo().renderingRuleInfos()` to get a list of `RenderingRuleInfo` supported by the service.
```

kotlin/raster-rendering-rule/README.md
```diff
@@ -1,34 +1,36 @@
...
+3.  Choose a rendering rule info to apply and use it to create a `RenderingRule`.
+4.  Create a new `ImageServiceRaster` using the same URL.
+5.  Apply the rendering rule to the new raster using `imageServiceRaster.renderingRule(renderingRuleInfo)`.
```
hud10837: Same as above
```suggestion
5.  Apply the rendering rule to the new raster using `ImageServiceRaster.renderingRule(renderingRuleInfo)`.
```

kotlin/scene-layer/README.md
```diff
@@ -14,12 +14,10 @@ When launched, this sample displays a scene service with an `ArcGISSceneLayer`.
 
 ## How it works
 
-1. Create an `ArcGISScene` and set its `Basemap` with `ArcGISScene.setBasemap()`.
-1. Create a `Surface` and add an elevation source to it: `surface.getElevationSources()add.(arcGISTiledElevationSource)`.
-1. Add the created surface to the scene: `ArcGISScene.setBaseSurface(surface)`.
+1. Create an `ArcGISScene` and set its `Basemap` with `ArcGISScene.basemap()`.
```
hud10837: I think it makes sense to get rid of the parentheses here.
```suggestion
1. Create an `ArcGISScene` and set its `Basemap` with `ArcGISScene.basemap`.
```

kotlin/scene-layer/README.md
```diff
@@ -14,12 +14,10 @@ When launched, this sample displays a scene service with an `ArcGISSceneLayer`.
...
 1. Create a `SceneView` and set the scene to the view with `SceneView.setArcGISScene(scene)`.
```
hud10837: Leftover setter here
```suggestion
1. Create a `SceneView` and set the scene to the view with `SceneView.scene`.
```

kotlin/scene-layer/README.md
```diff
@@ -14,12 +14,10 @@ When launched, this sample displays a scene service with an `ArcGISSceneLayer`.
...
-1. Create an `ArcGISSceneLayer` using a data source URI: `new ArcGISSceneLayer(Uri)`.
-1. Add the new scene layer to the scene as an operational layer with `ArcGISScene.getOperationalLayers().add(sceneLayer)`.
+1. Create an `ArcGISSceneLayer` using a data source URI: `ArcGISSceneLayer(Uri)`.
+1. Add the new scene layer to the scene as an operational layer with `ArcGISScene.operationalLayers().add(sceneLayer)`.
```
hud10837: Removing the parentheses here
```suggestion
1. Add the new scene layer to the scene as an operational layer with `ArcGISScene.operationalLayers.add(sceneLayer)`.
```

kotlin/scene-layer/README.md
```diff
@@ -31,8 +29,7 @@ The scene launches with a northward view of the city of Brest, France. A 3D scen
 * ArcGISSceneLayer
 * ArcGISTiledElevationSource
```
hud10837: `ArcGISTiledElevationSource` doesn't seem to be referenced here?

kotlin/show-labels-on-layer/README.md
```diff
@@ -1,26 +1,45 @@
...
+# Show labels on layers
 
-![Show Labels on Layer App](show-labels-on-layer.png)
+Display custom labels on a feature layer.
+
+![Image of show labels on layers](show-labels-on-layer.png)
+
+## Use case
+
+Labeling features is useful to visually display a key piece of information or attribute of a feature on a map. For example, you may want to label rivers or street with their names.
```
hud10837: street should be plural
```suggestion
Labeling features is useful to visually display a key piece of information or attribute of a feature on a map. For example, you may want to label rivers or streets with their names. 
```

kotlin/show-labels-on-layer/README.md
```diff
@@ -1,26 +1,45 @@
...
+To show custom labels on a feature layer:
+
+1. Create a `ServiceFeatureTable` using a feature service URL.
+2. Create a `FeatureLayer` from the service feature table.
+3. Create a `TextSymbol` to use for displaying the label text.
+4. Create a JSON string for the label definition.
+    * Set the "LabelExpressionInfo.expression" key to express what the text the label should display. You can use fields of the feature by using `$feature.field_name` in the expression.
+    * To use the text symbol, set the "symbol" key to the symbol's JSON representation using `textSymbol.toJson()`.
+5. Create a label definition from the JSON using `LabelDefinition.fromJson(json)`.
+6. Add the definition to the feature layer with `featureLayer.labelDefinitions().add(labelDefinition)` .
```
hud10837: remove parentheses
```suggestion
6. Add the definition to the feature layer with `featureLayer.labelDefinitions.add(labelDefinition)` .
```

kotlin/show-labels-on-layer/README.md
```diff
@@ -1,26 +1,45 @@
...
+7. Lastly, enable labels on the layer using `featureLayer.isLabelsEnabled()`.
```
hud10837: remove parentheses
```suggestion
7. Lastly, enable labels on the layer using `featureLayer.isLabelsEnabled`.
```

kotlin/wmts-layer/README.md
```diff
@@ -1,34 +1,42 @@
...
+![Image of WMTS layer](wmts-layer.png)
 
-## Features
+## Use case
+
+WMTS services can have several layers. You can use Runtime to explore the layers available from a service. This would commonly be used to enable a browsing experience where users can choose which layers they want to display at run time.
+
+## How to use the sample
+
+The layer will be displayed automatically. Use the buttons to choose a different method of loading the layer.
```
hud10837: It looks like there are no buttons on this sample, so you can probably cut the second sentence.

kotlin/raster-function-service/README.md
hud10837: Sorry, didn't realize this was kotlin at first, so I suppose this would be the same but with property access

kotlin/raster-rendering-rule/README.md
Rachael-E: good question: usually when we've referred to the instantiation of a class in a previous step, then it's ok to use the informal variable call in this case. That's probably something we could be a bit more consistent about, I'll keep an eye out for that with future readmes too.

kotlin/scene-layer/README.md
Rachael-E: ah yes, thanks: in the common design there is a Surface added which would use this class. But in our Android implementation we don't use a surface. I removed the references to that but forgot about this class! I'll update the Java one too.

kotlin/raster-rendering-rule/README.md
Rachael-E: yes there should, whoops!

kotlin/raster-function-service/README.md
Rachael-E: the main steps are captured above, but the common design has bunched two steps together. I'll separate out the getters of arguments and rasternames to be more readable.

Rachael-E:
thanks for the thorough review @hud10837! can you take another look please?

kotlin/raster-rendering-rule/README.md
hud10837: Okay I'll stop correcting those then! 😄

PR #719, opened 2020-02-21
Readmes rach kotlin2

Rachael-E:
Turns out the other ones I had planned for this one are already done! So just two readmes to look at here (just a few minor tweaks since these were recent samples). @hud10837 could you take the first review please?


Rachael-E:
great thanks! @sorenoid could you take a second review please?

PR #723, opened 2020-02-24
Readmes rach java3

Rachael-E:
Next batch of readmes ready for review. @sorenoid please could you take a first look when you have the opportunity? thanks!

- SceneLayerSelection
- ScenePropertiesExpression
- SceneSymbols
- SearchForWebmap
- ServiceFeatureTable cache

### DIFF
--- a/java/scene-property-expressions/README.md
+++ b/java/scene-property-expressions/README.md
@@ -15,21 +15,20 @@ Adjust the heading and pitch sliders to rotate the cone.
 ## How it works
 
 1. Create a new `GraphicsOverlay`. 
-2. Create a new `SimpleRenderer` and set its scene properties.
-3. Set the heading expression to `[HEADING]` and the pitch expression to `[PITCH]` with `renderer.getSceneProperties().setHeadingExpression`.
-4. Apply the renderer to the graphics overlay with `graphicsOverlay.setRenderer(renderer)`.
+2. Create a new `SimpleRenderer`.
+3. Set the heading expression to `[HEADING]` and the pitch expression to `[PITCH]` with `simpleRenderer.getSceneProperties().setHeadingExpression`.
+4. Apply the renderer to the graphics overlay with `graphicsOverlay.setRenderer(simpleRenderer)`.
 5. Create a new `Point` and a new `Graphic` and add it to the overlay with e.g. `graphicsOverlay.getGraphics().add(graphic)`.
 6. To update the graphic's rotation, update the `HEADING` or `PITCH` property in the graphic's attributes with `graphic.getAttributes().put(key, value)`.
        
 ## Relevant API
 
-* Graphic.Attributes
+* Graphic
 * GraphicsOverlay
 * SceneProperties
-* SceneProperties.HeadingExpression
-* SceneProperties.PitchExpression
+* SceneProperties.setHeadingExpression
+* SceneProperties.setPitchExpression
 * SimpleRenderer
-* SimpleRenderer.SceneProperties
 
 ## Tags
 
--- a/java/scene-property-expressions/README.md
+++ b/java/scene-property-expressions/README.md
@@ -2,53 +2,35 @@
 
 Update the orientation of a graphic using expressions based on its attributes.
 
-![Scene property expressions App](scene-property-expressions.png)
+![Image of scene property expressions](scene-property-expressions.png)
 
 ## Use case
 
-Instead of reading the attribute and changing the rotation on the symbol
-for a single graphic (a manual CPU operation), you can bind the rotation
-to an expression that applies to the whole overlay (an automatic GPU
-operation). This usually results in a noticeable performance boost
-(smooth rotations).
+Instead of reading the attribute and changing the rotation on the symbol for a single graphic (a manual CPU operation), you can bind the rotation to an expression that applies to the whole overlay (an automatic GPU operation). This usually results in a noticeable performance boost (smooth rotations).
 
 ## How to use the sample
 
-Use the seek bars to adjust the heading and pitch properties for the
-graphic.
+Adjust the heading and pitch sliders to rotate the cone.
 
 ## How it works
 
-Simple renderers can dynamically update the positions of graphics using
-an expression. The expression relates a renderer property to one of the
-graphic's attribute keys.
-
 1. Create a new `GraphicsOverlay`. 
 2. Create a new `SimpleRenderer` and set its scene properties.
-3. Set the heading expression to `[HEADING]` and the pitch expression to
-   `[PITCH]`.
-4. Apply the renderer to the graphics overlay.
-5. Create a new `Point` and a new `Graphic` and add it to the overlay.
-6. To update the graphic's rotation, update the `HEADING` or `PITCH`
-   property in the graphic's attributes.
+3. Set the heading expression to `[HEADING]` and the pitch expression to `[PITCH]` with `renderer.getSceneProperties().setHeadingExpression`.
+4. Apply the renderer to the graphics overlay with `graphicsOverlay.setRenderer(renderer)`.
+5. Create a new `Point` and a new `Graphic` and add it to the overlay with e.g. `graphicsOverlay.getGraphics().add(graphic)`.
+6. To update the graphic's rotation, update the `HEADING` or `PITCH` property in the graphic's attributes with `graphic.getAttributes().put(key, value)`.
        
 ## Relevant API
 
+* Graphic.Attributes
 * GraphicsOverlay
-* SimpleRenderer
-* SceneProperties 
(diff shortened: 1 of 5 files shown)


java/scene-symbols/README.md
```diff
@@ -2,25 +2,36 @@
...
+* Graphic
+* GraphicsOverlay
+* SceneSymbol.AnchorPosition
 * SimpleMarkerSceneSymbol
 * SimpleMarkerSceneSymbol.Style
-* SceneSymbol.AnchorPosition
+
+## About the data
+
+This sample shows arbitrary symbols in an empty scene with an  imagery basemap.
```
sorenoid: looks like there is an excess space between "an" and "imagery" ?

Rachael-E:
thanks @sorenoid! @hud10837 please could you do a second review, thanks.

java/scene-property-expressions/README.md
```diff
@@ -2,53 +2,35 @@
...
+Adjust the heading and pitch sliders to rotate the cone.
 
 ## How it works
 
-Simple renderers can dynamically update the positions of graphics using
-an expression. The expression relates a renderer property to one of the
-graphic's attribute keys.
-
 1. Create a new `GraphicsOverlay`. 
 2. Create a new `SimpleRenderer` and set its scene properties.
```
hud10837: This step seems redundant to me as it looks like the only properties that are set are the expressions, which are included in the next step

java/scene-property-expressions/README.md
```diff
@@ -2,53 +2,35 @@
...
-6. To update the graphic's rotation, update the `HEADING` or `PITCH`
-   property in the graphic's attributes.
+3. Set the heading expression to `[HEADING]` and the pitch expression to `[PITCH]` with `renderer.getSceneProperties().setHeadingExpression`.
+4. Apply the renderer to the graphics overlay with `graphicsOverlay.setRenderer(renderer)`.
+5. Create a new `Point` and a new `Graphic` and add it to the overlay with e.g. `graphicsOverlay.getGraphics().add(graphic)`.
+6. To update the graphic's rotation, update the `HEADING` or `PITCH` property in the graphic's attributes with `graphic.getAttributes().put(key, value)`.
        
 ## Relevant API
 
+* Graphic.Attributes
```
hud10837: Is this also meant to be just `Graphic`?

hud10837: Same for the ones below ⬇️

java/scene-symbols/README.md
```diff
@@ -2,25 +2,36 @@
...
+1. Create various simple marker scene symbols by specifying different styles and colors, and a height, width, depth, and anchor position of each.
+1. Create a new `Graphic` object for each symbol.
+1. Add the graphics to the graphics overlay with `graphicsOverlay.getGraphics().add(graphic)`.
+1. Add the graphics overlay to the scene view with `sceneView.getGraphicsOverlays().add(graphicsOverlay)`.
 
 ## Relevant API
 
+* Graphic
+* GraphicsOverlay
+* SceneSymbol.AnchorPosition
```
hud10837: I think these are also meant to be cut to just the class name

java/search-for-webmap/README.md
```diff
@@ -2,24 +2,23 @@
...
 
-Enter search terms into the search bar (use simple searches like "fire", "census", etc.). Once the search is complete a drawer will open listing the resultant webmaps. Tap on a webmap to set it to the map view. Scrolling to the bottom of the webmap recycler view will get more results.
+Enter search terms into the search bar. Once the search is complete, a list is populated with the resultant webmaps. Tap on a webmap to set it to the map view. Scrolling to the bottom of the webmap recycler view will get more results.
 
 ## How it works
 
 1. Create a new `Portal` and load it.
-2. Create new `PortalItemQueryParameters`. Set the type to `PortalItem.Type.WEBMAP` and the query a keyword you want to search.
-3. Use `portal.findItemsAsync(params)` to get the first set (10 by default) of matching items.
+2. Create new `PortalItemQueryParameters`. Set the type to `PortalItem.Type.WEBMAP` and add the text you want to search for.
```
hud10837: Looks like this class is called `PortalQueryParameters`
```suggestion
2. Create new `PortalQueryParameters`. Set the type to `PortalItem.Type.WEBMAP` and add the text you want to search for.
```

java/service-feature-table-cache/README.md
```diff
@@ -1,18 +1,37 @@
...
+![Image of service feature table on interaction cache](service-feature-table-cache.png)
+
+## Use case
+
+`ServiceFeatureTable` supports three request modes, which define how features are requested from the service and stored in the local table. The feature request modes have different performance characteristics. Use **On interaction cache** in scenarios with large amounts of infrequently edited data. See [Table performance concepts](https://developers.arcgis.com/net/latest/wpf/guide/layers.htm#ESRI_SECTION1_40F10593308A4718971C9A8F5FB9EC7D) to learn more.
 
 ## How to use the sample
-Simply run the app.
+
+Run the sample and pan and zoom around the map. With each interaction, features will be requested and stored in a local cache. Each subsequent interaction will display features from the cache and only request new features from the service.
```
hud10837: Just a note, don't think you need to change anything: it seems like this sample automatically sets its viewpoint on the extent of the data so panning and zooming doesn't actually introduce new data

java/service-feature-table-cache/README.md
```diff
@@ -1,18 +1,37 @@
...
 ## How it works
-Set the `ServiceFeatureTable.FeatureRequestMode` property of the service feature table to `ON_INTERACTION_CACHE` before the table is loaded. The mode cannot be changed once the table has been loaded.
+
+1. Set the `ServiceFeatureTable.FeatureRequestMode` property of the service feature table to `ON_INTERACTION_CACHE` before the table is loaded.
+2. Add the table to the map using a `FeatureLayer`; features will be requested for the visible extent as the user pans and zooms.
 
 ## Relevant API
+
 * FeatureLayer
+* FeatureRequestMode.OnInteractionCache
```
hud10837:
```suggestion
* FeatureRequestMode
```

java/service-feature-table-cache/README.md
```diff
@@ -1,18 +1,37 @@
...
 * ServiceFeatureTable
-* ServiceFeatureTable.setFeatureRequestMode(...)
+* ServiceFeatureTable.setFeatureRequestMode
```
hud10837:
```suggestion
```

java/scene-property-expressions/README.md
Rachael-E: ah, thanks! @sorenoid and I had a wee chat about use of methods in the Relevant API section at the end of the day yesterday. The summary of that discussion was that it turns out it is ok to refer to methods in this section, but that the Java/Android API will be more likely just to refer to classes in this section (it's a different story for other SDKs). 

However, we agreed if there was a particular method that we thought was deserving of being left in or directly relevant to the sample we could leave it in. This is where yourself and Soren's new exposure to our API will come in handy to judge if we should leave some methods in, if you think it adds to the readme. I think this one is a good example. Here, I'll remove Graphic.getAttribute since it's not pertinent to the sample, however I'll leave in SceneProperties.setHeading/PitchExpression since these are key for this sample. 

Going forward though I wouldn't add any methods to designs which don't already refer to them: and if they do, let's use our own judgement from a Java/Android perspective to assess if they add to the readme or not.

java/scene-symbols/README.md
Rachael-E: this is an enum within the class SceneSymbol, so is ok to leave in.

java/search-for-webmap/README.md
Rachael-E: nice catch!

java/service-feature-table-cache/README.md
Rachael-E: this is an enum, but I need to update it to look more like Java! Thanks!

java/service-feature-table-cache/README.md
Rachael-E: I think we can keep this in after the discussion above with Soren.

java/service-feature-table-cache/README.md
Rachael-E: good point: there's probably a sample enhancement idea in here, though the sample still does its job in showing the workflow and describing what is happening.

Rachael-E:
thanks for that @hud10837! Can you take another look please?

java/scene-property-expressions/README.md
hud10837: Oh no I saw this after going through one of Soren's PRs and it corrected all these 🙈 I'll keep this in mind.

PR #729, opened 2020-02-25
Readmes/densify to display drawing status

hud10837:
Hey @sorenoid these were pretty small, could you do a first review please?


hud10837:
Thanks @sorenoid does that look good now?

hud10837:
Thanks! @TADraeseke can you do a second review?

PR #741, opened 2020-02-27
Sor10874/readmes/56 58

sorenoid:
only three in this one.

picture-marker-symbol
play-a-kml-tour  (not in the Sample Viewer?)
portal-user-info


sorenoid:
wrong branch

PR #745, opened 2020-02-27
set android support versions to binary equivalent androidx versions. 

sorenoid:
same libraries, updated names. 
com.android.support:appcompat-v7 v28.0.0 is androidx.appcompat:appcompat v1.0.0
and 
com.android.support:design v28.0.0 is com.google.android.material:material v1.0.0

…see https://developer.android.com/jetpack/androidx/migrate/artifact-mappings

Testing across older android versions seems necessary. Sorry to say I am not quite sure how to go about that other than alerting Hardik, any thoughts @TADraeseke ?

### DIFF
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -26,8 +26,8 @@ subprojects{
     afterEvaluate {project ->
         if(project.hasProperty("dependencies")){
             dependencies {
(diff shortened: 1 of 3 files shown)


ZackAllen:
I am not familiar enough with the Android project or Gradle to give this a meaningful review.

PR #752, opened 2020-03-03
Show callout kotlin sample

Rachael-E:
This PR is for the kotlin sample of show callout. Straightforward, the only thing I changed from the Java version was how we set the text (I moved it to the strings.xml file instead of concatenating in the code). Hopefully this renders ok on other devices. @hud10837 could you take a first review please? Thanks!


kotlin/show-callout/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    // arcgis-android & appcompat lib dependencies from rootProject build.gradle
+    implementation "com.android.support.constraint:constraint-layout:$constraintLayoutVersion"
```
hud10837: Are we supposed to use the androidx dependency here? I'm not sure, it seems that both work fine. I'll put this here anyway.

```suggestion
    implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
```

Rachael-E: oh cool, thanks for pointing that out! I checked with Trevor, yeh you're right, we should: I've updated the wiki too!

Rachael-E:
@sorenoid could you take a second look please?

PR #756, opened 2020-03-09
Feature Layer Selection -> Kotlin 

Rachael-E:
This was a fairly straight forward conversion to Kotlin. I've opted for a `.let` on the `mapView` so as to keep it in scope for the DefaultMapViewOnTouchListener. When comparing it to the Java Android implementation, I also removed a Logcat for getting the feature table name since it's not part of the design and I don't think is required for the sample. I also slightly increased the tolerance so that more than one feature could be easily selected (so it's not just saying "1 feature selected" the whole time).

@TADraeseke could you take a first look please when you have a moment? Thanks!


kotlin/feature-layer-selection/src/main/java/com/esri/arcgisruntime/sample/featurelayerselection/MainActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+      // add the feature layer to the map's operational layers
+      operationalLayers.add(featureLayer)
+    }
+
+    mapView.let {
+      // set the map to be displayed in the layout's map view
+      it.map = streetsMap
+      // give any item selected on the map view a red selection halo
+      it.selectionProperties.color = Color.RED
+
```
TADraeseke: maybe remove whitespace here and 2 lines below to match compactness of the rest of the sample

kotlin/feature-layer-selection/src/main/java/com/esri/arcgisruntime/sample/featurelayerselection/MainActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+    val streetsMap = ArcGISMap(Basemap.createStreets()).apply {
+      // set an initial view point
+      initialViewpoint = Viewpoint(Envelope(
+        -1131596.019761, 3893114.069099, 3926705.982140, 7977912.461790,
+        SpatialReferences.getWebMercator()))
+      // add the feature layer to the map's operational layers
+      operationalLayers.add(featureLayer)
+    }
+
+    mapView.let {
```
TADraeseke: Fine to use a let, but apply is still possible here just FYI. For the onTouchListener you'd have to do:

it.onTouchListener = object : DefaultMapViewOnTouchListener(this@MainActivity, this) { (or something like this, I haven't got this loaded in a compiler at the moment)

Rachael-E: oh cool, that's good to know thanks! I just saw it complain with apply and assumed you couldn't use it since it didn't return a Unit? I think it was. Good to know the different parameters can get around this. I'll keep it as let in the meanwhile. Though I couldn't get the apply to compile using the above without recalling mapView explicitly and it seems yucky :(

Rachael-E:
@TADraeseke thanks! Those changes made. I'll ask @hud10837 to do a second review if you could approve those changes? Thanks!

kotlin/feature-layer-selection/README.md
```diff
@@ -0,0 +1,34 @@
+# Feature layer selection
+
+Select features in a feature layer.
+
+![Image of feature layer selection](feature-layer-selection.png)
+
+## Use case
+
+Selecting features, whether by query or identify, can be an important step both in editing data and visualizing results. One possible use case would be to query a feature layer containing street furniture. A query might look for type "bench" and return a list of bench features contained in the features with an attribute of type bench. These might be selected for further editing (see FeatureQueryResult) or may just be highlighted visually.
```
hud10837: "whether by query or identify" probably is worded correctly according to api, but it sounds pretty strange. Perhaps identification?

kotlin/feature-layer-selection/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    // arcgis-android & appcompat lib dependencies from rootProject build.gradle
+    implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
```
hud10837: Something Soren has been pointing out: We don't really need the constraint layout if it's just the one map view, and you could remove it and the dependency. But we should keep the comment on line 25 and the dependencies block to remind people about arcgis dependencies

kotlin/feature-layer-selection/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="com.esri.arcgisruntime.sample.featurelayerselection.MainActivity">
+
+    <com.esri.arcgisruntime.mapping.view.MapView
```
hud10837: As mentioned above, it's possible to remove the constraint layout and just have the mapview here

kotlin/feature-layer-selection/README.md
Rachael-E: I know what you mean, but since it makes sense from an API perspective, let's leave as is as I think it's clear enough :)

Rachael-E:
thanks @hud10837, I removed the constraint layout! Merging now.

PR #757, opened 2020-03-10
Trev8939/provision

TADraeseke:
Hi @Rachael-E - Could you review please?


java/display-kml/README.metadata.json
```diff
@@ -12,6 +12,12 @@
         "Portal",
         "PortalItem"
     ],
+    "provision_from": [
+        "https://arcgisruntime.maps.arcgis.com/home/item.html?id=324e4742820e46cfbe5029ff2c32cb1f"
+    ],
+    "provision_to": [
+        "/US_State_Capitals.kml"
```
Rachael-E: looking at the path for this data in the sample, I think this should be `/ArcGIS/Samples/KML/US_State_Capitals.kml`?

java/generate-geodatabase/README.metadata.json
```diff
@@ -15,6 +15,12 @@
         "MapView",
         "ServiceFeatureTable"
     ],
+    "provision_from": [
+        "https://arcgisruntime.maps.arcgis.com/home/item.html?id=72e703cd01654e7796eb1ae75af1cb53"
+    ],
+    "provision_to": [
+        "/ArcGIS/samples/TileCache/SanFrancisco.tpk"
```
Rachael-E: not sure how much casing matters but, wondering if this should be `/Samples/` to match the resource string in the sample. Same applies for other instances of this in the PR

kotlin/custom-dictionary-style/README.metadata.json
```diff
@@ -13,6 +13,12 @@
         "Style",
         "Unique value"
     ],
+    "provision_from": [
+        "https://arcgisruntime.maps.arcgis.com/home/item.html?id=751138a2e0844e06853522d54103222a"
+    ],
+    "provision_to": [
+        "/Restaurant.stylx"
```
Rachael-E: just a question, does it matter that the data will be stored in two different ways? e.g. this way vs storing it in an ArcGIS/Samples directory. I wondered if the former is legacy from using SD card storage? And going forward with Kotlin we'd keep with the "/dataString.tpk" format?

java/generate-geodatabase/README.metadata.json
TADraeseke: I think it's ok, but thanks for checking!

kotlin/custom-dictionary-style/README.metadata.json
TADraeseke: Eagle eyes! Yes, alot of the longer preamble in the paths ArcGIS/Samples/<data_type>/ will be dropped as we move away from SD card storage towards scoped storage

TADraeseke:
Thanks @Rachael-E!

PR #772, opened 2020-03-20
Missing readmes

hud10837:
Hiya, these are the missing readmes that Rachael mentioned in the readme PR. Most of them were actually here already, so pay close attention to "Unique value renderer" and "feature layer show attributes", which are new. Thanks!


java/feature-layer-show-attributes/README.md
```diff
@@ -1,19 +1,30 @@
-# Feature Layer Show Attributes
+# Feature layer show attributes
+
 Return all loaded features from a query to show all attributes.
 
 ![Feature Layer Show Attributes App](feature-layer-show-attributes.png)
 
+## Use case
+
+Attributes can be used to help describe the objects represented by a feature.
```
TADraeseke:
```suggestion
Attributes can be used to help describe the objects represented by a feature. A geologist might use an attribute to describe the rock type of polygons representing surface geology. Archaeologists might use attributes to record the stratigraphic layer of finds represented as point features.
```

or something like that

hud10837:
@TADraeseke thanks for the suggestion. Unfortunately this is the sample without a common design or implementation on other platforms---I don't know if you saw the chat.

PR #778, opened 2020-03-31
Fixes for verification

hud10837:
Hi @TADraeseke , I caught an unfinished branch for the changes needed on some of the samples for verification. I'm not sure if you caught these on your own, but could you double check I've done the offline data section correctly if not? Thanks


kotlin/offline-routing/README.md
```diff
@@ -26,17 +26,15 @@ Tap once near a road to add a stop to the map. A number graphic will show its or
 1. Download the data from [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=567e14f3420d40c5a206e5c0284cf8fc).
 2. Extract the contents of the downloaded zip file to disk.
 3. Open your command prompt and navigate to the folder where you extracted the contents of the data from step 1.
-4. Execute the following commands:
+4. Make a `/sandiego.tn/` directory for the `sandiego.tn` file within the scoped storage of the sample app.
```
TADraeseke, line 29: I don't think you need to make the directory. You can just push a directory if needed (ie push san_diego)

hud10837: You're right, I discovered that after making this change, will fix.

hud10837:
@TADraeseke I've made both of those changes now! For the error, I just added a single line at the start of the updateRoute() function

```cs
    // do not calculate a route if there is only one stop
    if (stops.size <= 1) return
```

kotlin/offline-routing/src/main/java/com/esri/arcgisruntime/sample/offlinerouting/MainActivity.kt
```diff
@@ -182,6 +182,8 @@ class MainActivity : AppCompatActivity() {
     val stops = stopsOverlay.graphics.map {
       Stop(it.geometry as Point)
     }
+    // do not calculate a route if there is only one stop
+    if (stops.size <= 1) return
```
TADraeseke, line 186: I've copied this into the 1870fixes PR, so this can be removed

kotlin/offline-routing/README.md
```diff
@@ -26,17 +26,14 @@ Tap once near a road to add a stop to the map. A number graphic will show its or
 1. Download the data from [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=567e14f3420d40c5a206e5c0284cf8fc).
 2. Extract the contents of the downloaded zip file to disk.
 3. Open your command prompt and navigate to the folder where you extracted the contents of the data from step 1.
-4. Execute the following commands:
```
TADraeseke, line 29: I think this is handled in the big android Q update. If so, can we remove changes to this file?

hud10837:
@TADraeseke I hope I've managed that correctly!

PR #779, opened 2020-03-31
Remove isolated trace

hud10837:
This is a tiny fix to address #1768 by replacing the generated array of trace types with a hardcoded one. @TADraeseke could you take a quick look?


hud10837:
Thanks Trevor!

PR #792, opened 2020-05-06
add nullability check to dialog onclick

hud10837:
Hello @TADraeseke, this is just to solve the build errors for the sample viewer related to this sample. It seems to be building the sample viewer fine now. Could you have a go and take a look? Thanks.


TADraeseke:
I still hit a build build failure...it's wanting to change onDismiss(...) to dismiss()

TADraeseke:
This was addressed as part of the fix we discussed. Closing :)

PR #800, opened 2020-05-20
Hud10837/offline geocode kotlin

hud10837:
This is the offline geocode sample conversion. It's not too different from the java, but I tried to implement a search view in the same style as the find address sample. Also, I changed the behavior of real-time reverse geocoding; in the java version, the user had to select and then tap+drag, but to simplify it I made it a double-touch-drag gesture. @TADraeseke can you take a look when you get the chance?


kotlin/offline-geocode/src/main/java/com/esri/arcgisruntime/sample/offlinegeocode/MainActivity.kt
```diff
@@ -0,0 +1,278 @@
...
+      suggestionsAdapter = SimpleCursorAdapter(
+        this@MainActivity,
+        R.layout.suggestion,
+        suggestionsCursor,
+        arrayOf("address"),
+        intArrayOf(R.id.suggestion_address),
+        0
+      )
+      // show the suggestions as soon as the user opens the search view
+      findViewById<AutoCompleteTextView>(R.id.search_src_text).threshold = 0
```
TADraeseke, line 255: I see that this doesn't work with kotlin synthetic. Any idea why?

kotlin/offline-geocode/src/main/java/com/esri/arcgisruntime/sample/offlinegeocode/MainActivity.kt
```diff
@@ -0,0 +1,278 @@
...
+        override fun onSuggestionSelect(position: Int) = true
+
+        override fun onSuggestionClick(position: Int): Boolean {
+          geocode(suggestions[position])
+          return true
+        }
+      })
+    }
+  }
+}
```
TADraeseke, line 278: missing life cycle overrides

kotlin/offline-geocode/src/main/java/com/esri/arcgisruntime/sample/offlinegeocode/MainActivity.kt
hud10837: So this line is meant to fix the problem you mentioned above, which is that the suggestions should appear when the search view is opened. This behavior works for me, so I'm curious if it doesn't work for you

hud10837: The reason this does not use kotlin synthetic is because this view is not directly part of activity_main. Because this is in an `apply` block, this is really doing a `searchView.findViewById<...>(...)`. The `R.id.search_src_text` is an android-defined id, and under the hood, the searchView contains an `AutoCompleteTextView`. I can't remember what the more complicated solution for this behavior was, but I remember struggling for a long time trying to figure out how to implement it succinctly. Basically, this line is unpacking the searchView, grabbing the view that controls the suggestions, and setting its threshold for auto-completing to 0 so that the user doesn't have to type any characters to display suggestions

hud10837:
@TADraeseke Thanks for this review! I've made all of the changes you suggested and addressed your question about the findViewById, which also covers the autocomplete suggestions. If you are not getting the autocomplete suggestions when you first tap the searchview, then I think I will have to change this sample to include a spinner. Should I go ahead and do that?

PR #812, opened 2020-06-08
Hud10837/garbage collection fixes

hud10837:
This PR covers the issue with Loadables being garbage collected on Android in both java and kotlin (#1886). The changes made here were to move any loadables created within the local scope of a function out of the local scope and into the scope of the class. On Java, this was fairly straightforward; on kotlin, I tried to evaluated case-by-case what type of val or var the class field should be.

I left comments next to the member fields that I added. However, these may be inconsistent given that some loadables were already class-level and some loadables were not moved to the class-level because they were added (indirectly or directly) to a geoview. 

@TADraeseke can you take a first look?

### DIFF
--- a/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
+++ b/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
@@ -55,6 +55,7 @@ public class MainActivity extends AppCompatActivity {
   private GraphicsOverlay mSceneOverlay;
   private OrbitGeoElementCameraController mOrbitPlaneCameraController;
   private OrbitLocationCameraController mOrbitLocationCameraController;
+  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
   private ModelSceneSymbol mModelSceneSymbol;
 
   @Override
--- a/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
+++ b/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
@@ -55,8 +55,6 @@ public class MainActivity extends AppCompatActivity {
   private GraphicsOverlay mSceneOverlay;
   private OrbitGeoElementCameraController mOrbitPlaneCameraController;
   private OrbitLocationCameraController mOrbitLocationCameraController;
-  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
-  private ModelSceneSymbol mModelSceneSymbol;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
@@ -96,8 +94,7 @@ public class MainActivity extends AppCompatActivity {
     mOrbitLocationCameraController.setCameraPitchOffset(3);
     mOrbitLocationCameraController.setCameraHeadingOffset(150);
 
-    mModelSceneSymbol = loadModel();
-    mModelSceneSymbol.addDoneLoadingListener(() -> {
+    loadModel().addDoneLoadingListener(() -> {
       // instantiate a new camera controller which orbits the plane at a set distance
       mOrbitPlaneCameraController = new OrbitGeoElementCameraController(mPlane3D, 100.0);
       mOrbitPlaneCameraController.setCameraPitchOffset(30);
@@ -138,7 +135,6 @@ public class MainActivity extends AppCompatActivity {
     // create a graphic with a ModelSceneSymbol of a plane to add to the scene
     String pathToModel = getCacheDir() + File.separator + getString(R.string.bristol_model);
     ModelSceneSymbol plane3DSymbol = new ModelSceneSymbol(pathToModel, 1.0);
-    plane3DSymbol.loadAsync();
     plane3DSymbol.setHeading(45);
     mPlane3D = new Graphic(new Point(-109.937516, 38.456714, 5000, SpatialReferences.getWgs84()), plane3DSymbol);
     mSceneOverlay.getGraphics().add(mPlane3D);
--- a/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
+++ b/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
@@ -55,6 +55,7 @@ public class MainActivity extends AppCompatActivity {
   private GraphicsOverlay mSceneOverlay;
   private OrbitGeoElementCameraController mOrbitPlaneCameraController;
   private OrbitLocationCameraController mOrbitLocationCameraController;
+  private ModelSceneSymbol mModelSceneSymbol;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
@@ -94,7 +95,8 @@ public class MainActivity extends AppCompatActivity {
     mOrbitLocationCameraController.setCameraPitchOffset(3);
     mOrbitLocationCameraController.setCameraHeadingOffset(150);
 
-    loadModel().addDoneLoadingListener(() -> {
+    mModelSceneSymbol = loadModel();
+    mModelSceneSymbol.addDoneLoadingListener(() -> {
       // instantiate a new camera controller which orbits the plane at a set distance
       mOrbitPlaneCameraController = new OrbitGeoElementCameraController(mPlane3D, 100.0);
       mOrbitPlaneCameraController.setCameraPitchOffset(30);
--- a/java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
+++ b/java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
@@ -34,8 +34,6 @@ public class MainActivity extends AppCompatActivity {
   private final static String TAG = MainActivity.class.getSimpleName();
 
   private MapView mMapView;
-  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
-  private ShapefileFeatureTable mShapefileFeatureTable;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
@@ -48,22 +46,20 @@ public class MainActivity extends AppCompatActivity {
     mMapView.setMap(map);
 
     // load the shapefile with a local path
-    mShapefileFeatureTable = new ShapefileFeatureTable(
+    ShapefileFeatureTable shapefileFeatureTable = new ShapefileFeatureTable(
         getExternalFilesDir(null) + getString(R.string.shapefile_path));
 
     // create a feature layer to display the shapefile
-    FeatureLayer shapefileFeatureLayer = new FeatureLayer(mShapefileFeatureTable);
+    FeatureLayer shapefileFeatureLayer = new FeatureLayer(shapefileFeatureTable);
+    // add the feature layer to the map
+    mMapView.getMap().getOperationalLayers().add(shapefileFeatureLayer);
 
-    mShapefileFeatureTable.addDoneLoadingListener(() -> {
-      if (mShapefileFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
-
-        // add the feature layer to the map
-        mMapView.getMap().getOperationalLayers().add(shapefileFeatureLayer);
-
+    shapefileFeatureTable.addDoneLoadingListener(() -> {
+      if (shapefileFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
         // zoom the map to the extent of the shapefile
         mMapView.setViewpointAsync(new Viewpoint(shapefileFeatureLayer.getFullExtent()));
       } else {
-        String error = "Shapefile feature table failed to load: " + mShapefileFeatureTable.getLoadError().toString();
+        String error = "Shapefile feature table failed to load: " + shapefileFeatureTable.getLoadError().toString();
         Toast.makeText(MainActivity.this, error, Toast.LENGTH_LONG).show();
         Log.e(TAG, error);
       }
--- a/java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
+++ b/java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
@@ -34,6 +34,7 @@ public class MainActivity extends AppCompatActivity {
   private final static String TAG = MainActivity.class.getSimpleName();
 
   private MapView mMapView;
+  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
   private ShapefileFeatureTable mShapefileFeatureTable;
 
   @Override
--- a/java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
+++ b/java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
@@ -51,12 +51,11 @@ public class MainActivity extends AppCompatActivity {
     mShapefileFeatureTable = new ShapefileFeatureTable(
         getExternalFilesDir(null) + getString(R.string.shapefile_path));
 
-    mShapefileFeatureTable.loadAsync();
+    // create a feature layer to display the shapefile
+    FeatureLayer shapefileFeatureLayer = new FeatureLayer(mShapefileFeatureTable);
+
     mShapefileFeatureTable.addDoneLoadingListener(() -> {
       if (mShapefileFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
-
-        // create a feature layer to display the shapefile
-        FeatureLayer shapefileFeatureLayer = new FeatureLayer(mShapefileFeatureTable);
 
         // add the feature layer to the map
         mMapView.getMap().getOperationalLayers().add(shapefileFeatureLayer);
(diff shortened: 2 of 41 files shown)


hud10837:
@tschie I'm tagging you in if you could take a quick look at the kotlin samples when you get the chance! Thanks

hud10837:
Thanks Tyler!

java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
```diff
@@ -46,23 +48,23 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    shapefileFeatureTable.loadAsync();
-    shapefileFeatureTable.addDoneLoadingListener(() -> {
-      if (shapefileFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
+    mShapefileFeatureTable.loadAsync();
+    mShapefileFeatureTable.addDoneLoadingListener(() -> {
+      if (mShapefileFeatureTable.getLoadStatus() == LoadStatus.LOADED) {
 
         // create a feature layer to display the shapefile
-        FeatureLayer shapefileFeatureLayer = new FeatureLayer(shapefileFeatureTable);
+        FeatureLayer shapefileFeatureLayer = new FeatureLayer(mShapefileFeatureTable);
```
TADraeseke, line 59: I think we can just move this out of the done loading listener, which would keep the table in scope

java/find-route/src/main/java/com/esri/arcgisruntime/sample/findroute/MainActivity.java
```diff
@@ -233,41 +237,39 @@ private void setupSymbols() {
...
         public void run() {
           //add a new graphic as start point
           mSourcePoint = new Point(-117.15083257944445, 32.741123367963446, SpatialReferences.getWgs84());
-          Graphic pinSourceGraphic = new Graphic(mSourcePoint, pinSourceSymbol);
+          Graphic pinSourceGraphic = new Graphic(mSourcePoint, mPinSourceSymbol);
           mGraphicsOverlay.getGraphics().add(pinSourceGraphic);
         }
       });
-      pinSourceSymbol.setOffsetY(20);
+      mPinSourceSymbol.setOffsetY(20);
```
TADraeseke, line 252: I think this is another case where loadable chaining would work.

something like

```
pinSourceSymbol = PictureMarkerSymbol.createAsync(startDrawable).get();
      pinSourceSymbol.setOffsetY(20);
      //add a new graphic as start point
      mSourcePoint = new Point(-117.15083257944445, 32.741123367963446, SpatialReferences.getWgs84());
      Graphic pinSourceGraphic = new Graphic(mSourcePoint, pinSourceSymbol);
      mGraphicsOverlay.getGraphics().add(pinSourceGraphic);
    } catch (InterruptedException | ExecutionException e) {
      String error = "Error creating picture marker symbol: " + e.getMessage();
      Log.e(TAG, error);
      Toast.makeText(this, error, Toast.LENGTH_SHORT).show();
```

java/picture-marker-symbols/src/main/java/com/esri/arcgisruntime/sample/picturemarkersymbols/MainActivity.java
```diff
@@ -81,44 +86,44 @@ protected void onCreate(Bundle savedInstanceState) {
...
     //its appearance would then differ across devices with different resolutions.
-    campsiteSymbol.setHeight(18);
-    campsiteSymbol.setWidth(18);
-    campsiteSymbol.loadAsync();
+    mCampsiteSymbol.setHeight(18);
+    mCampsiteSymbol.setWidth(18);
+    mCampsiteSymbol.loadAsync();
     //[DocRef: END]
-    campsiteSymbol.addDoneLoadingListener(new Runnable() {
+    mCampsiteSymbol.addDoneLoadingListener(new Runnable() {
```
TADraeseke, line 98: as above, I think we can rely on loadable chain here and therefore avoid another global property

java/style-wms-layer/src/main/java/com/esri/arcgisruntime/stylewmslayer/MainActivity.java
```diff
@@ -49,34 +51,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    wmsLayer.loadAsync();
-    wmsLayer.addDoneLoadingListener(() -> {
-      if (wmsLayer.getLoadStatus() == LoadStatus.LOADED) {
+    mWmsLayer = new WmsLayer(getString(R.string.wms_layer_url_minnesota), wmsLayerNames);
+    mWmsLayer.loadAsync();
+    mWmsLayer.addDoneLoadingListener(() -> {
+      if (mWmsLayer.getLoadStatus() == LoadStatus.LOADED) {
         // add the layer to the map
-        map.getOperationalLayers().add(wmsLayer);
+        map.getOperationalLayers().add(mWmsLayer);
```
TADraeseke, line 59: if we move this outside the doneLoading listener, it should be kept in scope

java/view-point-cloud-data-offline/src/main/java/com/esri/arcgisruntime/sample/viewpointclouddataoffline/MainActivity.java
```diff
@@ -59,24 +61,24 @@
...
         getExternalFilesDir(null) + getString(R.string.scene_layer_package_location));
 
     // add a listener to perform operations when the load status of the PointCloudLayer changes
-    pointCloudLayer.addDoneLoadingListener(() -> {
-      if (pointCloudLayer.getLoadStatus() == LoadStatus.LOADED) {
+    mPointCloudLayer.addDoneLoadingListener(() -> {
+      if (mPointCloudLayer.getLoadStatus() == LoadStatus.LOADED) {
         // add the PointCloudLayer to the operational layers of the scene
-        mSceneView.getScene().getOperationalLayers().add(pointCloudLayer);
+        mSceneView.getScene().getOperationalLayers().add(mPointCloudLayer);
```
TADraeseke, line 71: let's just add it to the scene outside of the done loading listener to keep it in scope

java/web-tiledlayer/src/main/java/com/esri/arcgisruntime/sample/webtiledlayer/MainActivity.java
```diff
@@ -44,15 +46,15 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    webTiledLayer.loadAsync();
-    webTiledLayer.addDoneLoadingListener(() -> {
-      if (webTiledLayer.getLoadStatus() == LoadStatus.LOADED) {
+    mWebTiledLayer = new WebTiledLayer(getString(R.string.template_uri_stamen), subDomains);
+    mWebTiledLayer.loadAsync();
+    mWebTiledLayer.addDoneLoadingListener(() -> {
+      if (mWebTiledLayer.getLoadStatus() == LoadStatus.LOADED) {
         // use web tiled layer as Basemap
-        ArcGISMap map = new ArcGISMap(new Basemap(webTiledLayer));
+        ArcGISMap map = new ArcGISMap(new Basemap(mWebTiledLayer));
```
TADraeseke, line 54: Move this outside of done loading listener to keep in scope instead of global

kotlin/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.kt
```diff
@@ -49,6 +49,8 @@ class MainActivity : AppCompatActivity() {
   private lateinit var plane3d: Graphic
   private lateinit var orbitLocationCameraController: OrbitLocationCameraController
   private lateinit var orbitPlaneCameraController: OrbitGeoElementCameraController
+  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
+  private val modelSceneSymbol: ModelSceneSymbol by lazy { loadModel() }
```
TADraeseke, line 53: not required since symbol is put on graphic and graphic onto graphics overlay outside of a done loading listener

kotlin/view-point-cloud-data-offline/src/main/java/com/esri/arcgisruntime/samples/viewpointclouddataoffline/MainActivity.kt
```diff
@@ -56,23 +56,11 @@ class MainActivity : AppCompatActivity() {
...
-        sceneView.scene.operationalLayers.add(pointCloudLayer)
-      } else {
-        val error = "Error loading point cloud layer: " + pointCloudLayer.loadError
-        Toast.makeText(this, error, Toast.LENGTH_LONG).show()
-        Log.e(TAG, error)
-      }
-    }
+    // add the PointCloudLayer to the scene
+    sceneView.scene.operationalLayers.add(pointCloudLayer)
   }
```
TADraeseke, line 64: I think we should keep the doneloading listener to catch errors (but add the pointCloudLayer like you're doing outside the done loading listener).

I vaguely feel like I might have said otherwise at some point--apologies if so!

java/view-content-beneath-the-terrain-surface/src/main/java/com/esri/arcgisruntime/sample/viewcontentbeneaththeterrainsurface/MainActivity.java
```diff
@@ -34,6 +34,10 @@
   private static final String TAG = MainActivity.class.getSimpleName();
 
   private SceneView mSceneView;
+  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
+  private Portal mPortal;
+  private PortalItem mPortalItem;
+  private ArcGISScene mScene;
```
TADraeseke, line 40: Let's just use the chainable loading to avoid all these.

Can put error catches in separate listeners where required. Probably just the portal item would suffice?

java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
```diff
@@ -94,7 +96,8 @@ protected void onCreate(Bundle savedInstanceState) {
     mOrbitLocationCameraController.setCameraPitchOffset(3);
     mOrbitLocationCameraController.setCameraHeadingOffset(150);
 
-    loadModel().addDoneLoadingListener(() -> {
+    mModelSceneSymbol = loadModel();
```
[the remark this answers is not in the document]
colinanderson: I agree with @TADraeseke. Also I don't think you need the `loadAsync()` here - https://github.com/Esri/arcgis-runtime-samples-android/blob/f1ab6aaa5f0e55a50e2ed056c6027c0ab9253d5e/java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java#L141 as the graphic should load it automatically.

java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
colinanderson: Yes, I think that is better. Also if you are adding the table to a layer I think you don't need to call `loadAsync()` on the table.

java/picture-marker-symbols/src/main/java/com/esri/arcgisruntime/sample/picturemarkersymbols/MainActivity.java
colinanderson: Yes, it looks like a lot of the samples do the pattern you have here i.e. load something then when loading is done add it to something that would have loaded it automatically anyway. It is worth remembering you can still have the done loading listeners to catch an error even if you use automatic loading.

TADraeseke:
Thanks @colinanderson! @hud10837 - I think we're good to go ahead with those proposed changes. Happy to help with any of them :)

java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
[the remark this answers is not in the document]
hud10837: I can't believe I missed that one! Thanks for catching both of those guys!

java/find-route/src/main/java/com/esri/arcgisruntime/sample/findroute/MainActivity.java
hud10837: this is a really nice refactor for readability as well!

hud10837:
Thanks @colinanderson  and @TADraeseke !! I can't believe I missed so many of those, I guess I was scared to remove any listeners that were already present. Everything looks much cleaner now! I've made all the changes, @TADraeseke  could you take another look?

java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
```diff
@@ -34,6 +34,8 @@
   private final static String TAG = MainActivity.class.getSimpleName();
 
   private MapView mMapView;
+  // objects that implement Loadable must be class fields to prevent being garbage collected before loading
+  private ShapefileFeatureTable mShapefileFeatureTable;
```
TADraeseke, line 38: This doesn't have to be a member variable because it's attached to the feature layer outside of a done loading listener

java/choose-camera-controller/src/main/java/com/esri/arcgisruntime/sample/choosecameracontroller/MainActivity.java
```diff
@@ -135,7 +135,6 @@ private ModelSceneSymbol loadModel() {
     // create a graphic with a ModelSceneSymbol of a plane to add to the scene
     String pathToModel = getCacheDir() + File.separator + getString(R.string.bristol_model);
     ModelSceneSymbol plane3DSymbol = new ModelSceneSymbol(pathToModel, 1.0);
-    plane3DSymbol.loadAsync();
```
Rachael-E, line 138: I didn't know this step wasn't necessary!

java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
Rachael-E: I'm unsure here: it looks to me that there's a reference to the feature table inside the done loading listener, in that if the scope were to be lost the user might not get an error back from it on line 66?

java/integrated-windows-authentication/src/main/java/com/esri/arcgisruntime/sample/integratedwindowsauthentication/MainActivity.java
```diff
@@ -85,11 +87,13 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
     // set up search public button
     Button searchPublicButton = findViewById(R.id.searchPublicButton);
     searchPublicButton.setOnClickListener(v -> {
       // search the the public ArcGIS portal
-      searchPortal(new Portal(getString(R.string.arcgis_url)));
+      searchPortal(mPortal);
     });
 
     // get reference to load state UI elements
```
Rachael-E, line 99: on line 112 you might want to make that new Portal that is being passed into the `searchPortal` method a class member too, since it will go through the addDone LoadingListener task too.

java/view-content-beneath-the-terrain-surface/src/main/java/com/esri/arcgisruntime/sample/viewcontentbeneaththeterrainsurface/MainActivity.java
```diff
@@ -43,35 +43,23 @@ protected void onCreate(Bundle savedInstanceState) {
...
-            String error = "Portal item failed to load: " + portalItem.getLoadError().getMessage();
-            Toast.makeText(this, error, Toast.LENGTH_LONG).show();
-            Log.e(TAG, error);
-          }
-        });
-        portalItem.loadAsync();
-      } else {
-        String error = "Portal failed to load: " + portal.getLoadError().getMessage();
-        Toast.makeText(this, error, Toast.LENGTH_LONG).show();
+    PortalItem portalItem = new PortalItem(portal, getString(R.string.subsurface_item_id));
```
Rachael-E, line 46: is there a reason this hasn't been made a class member?

java/feature-layer-shapefile/src/main/java/com/esri/arcgisruntime/sample/featurelayershapefile/MainActivity.java
TADraeseke: Oh yeah sorry, these lines should be outside of the doneLoadingListener:

```
 // add the feature layer to the map
        mMapView.getMap().getOperationalLayers().add(shapefileFeatureLayer);
```

Then the table doesn't have to be global, I believe

hud10837: my bad that was sloppy of me

java/integrated-windows-authentication/src/main/java/com/esri/arcgisruntime/sample/integratedwindowsauthentication/MainActivity.java
hud10837: good catch!

java/view-content-beneath-the-terrain-surface/src/main/java/com/esri/arcgisruntime/sample/viewcontentbeneaththeterrainsurface/MainActivity.java
hud10837: This portal item is later used to create a scene

hud10837:
@Rachael-E @TADraeseke  thanks for those comments! I've made those as requested.

hud10837:
sorry @tschie I did not mean to rerequest a review from you!

hud10837:
Thanks so much everyone! Merging now.

PR #819, opened 2020-06-12
Merge master into develop

hud10837:
@TADraeseke This also fixes the metadata for feature collection layer portal item, but only for the develop branch


kotlin/feature-collection-layer-portal-item/README.metadata.json
```diff
@@ -0,0 +1,32 @@
...
+        "feature collection layer",
+        "id",
+        "item",
+        "map notes",
+        "portal"
+    ],
+    "language": "kotlin",
+    "provision_from": "TODO",
+    "provision_to": "TODO",
+    "redirect_from": "TODO",
```
TADraeseke, line 21: These should be removed :)

PR #820, opened 2020-06-12
Trev8939/rename layer samples

TADraeseke:
Fixes name conflicts between samples, files and metadata to fix bug in the sample viewer


TADraeseke:
Merging to fix daily builds

PR #821, opened 2020-06-15
set cancelOnErrors to true on job to catch cancel from auth

TADraeseke:
Fix for @hardikc07's reported behaviour of job running after cancelling authentication challenge.

### DIFF
--- a/kotlin/generate-offline-map/src/main/java/com/esri/arcgisruntime/sample/generateofflinemap/MainActivity.kt
+++ b/kotlin/generate-offline-map/src/main/java/com/esri/arcgisruntime/sample/generateofflinemap/MainActivity.kt
@@ -100,8 +100,6 @@ class MainActivity : AppCompatActivity() {
       // update the download area box whenever the viewpoint changes
       addViewpointChangedListener {
         if (map.loadStatus == LoadStatus.LOADED) {
-          // enable the map offline button only after the map is loaded
-          if (!takeMapOfflineButton.isEnabled) takeMapOfflineButton.isEnabled = true
           // upper left corner of the area to take offline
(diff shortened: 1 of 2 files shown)


kotlin/generate-offline-map/src/main/java/com/esri/arcgisruntime/sample/generateofflinemap/MainActivity.kt
```diff
@@ -116,8 +114,10 @@ class MainActivity : AppCompatActivity() {
           if (minPoint != null && maxPoint != null) {
             val envelope = Envelope(minPoint, maxPoint)
             downloadArea.geometry = envelope
+            // enable the map offline button only after the map is loaded
+            if (!takeMapOfflineButton.isEnabled) takeMapOfflineButton.isEnabled = true
+          }
           }
```
hud10837, line 120: strange indent, but it looks like this is the only thing the reformatter picks up.

kotlin/generate-offline-map/src/main/java/com/esri/arcgisruntime/sample/generateofflinemap/MainActivity.kt
```diff
@@ -116,8 +114,10 @@ class MainActivity : AppCompatActivity() {
           if (minPoint != null && maxPoint != null) {
             val envelope = Envelope(minPoint, maxPoint)
             downloadArea.geometry = envelope
+            // enable the map offline button only after the map is loaded
```
hud10837, line 117:
```suggestion
            // enable the take map offline button only after the map is loaded
```

TADraeseke:
Thanks @hud10837 - Made those changes and merging now!

PR #823, opened 2020-06-17
Hud10837/convex hull

hud10837:
This is the convex hull implementation, a first for Android. I've also implemented the change to fix the wraparound bug. @TADraeseke could you take a look when you get the chance?


TADraeseke:
Merged that branch in, so all LGTM! Thanks @hud10837!

hud10837:
Thanks Trevor! Could you approve it so I can merge?

PR #830, opened 2020-06-22
Hud10837/fixes for sample viewer release

hud10837:
This just changes the show location history behavior so that when the user has panned away, pressing the button turns on the auto pan mode again. I also changed the behavior of attribution view change to show a toast instead of a snackbar and to change position directly instead of animate

### DIFF
--- a/kotlin/show-location-history/src/main/java/com/esri/arcgisruntime/sample/showlocationhistory/MainActivity.kt
+++ b/kotlin/show-location-history/src/main/java/com/esri/arcgisruntime/sample/showlocationhistory/MainActivity.kt
@@ -102,8 +102,14 @@ class MainActivity : AppCompatActivity() {
       initialZoomScale = 7000.0
     }
 
-    // change the isTrackLocation flag and the button's icon
+    // reset location display and change the isTrackLocation flag and the button's icon
     button.setOnClickListener {
+      // if the user has panned away from the location display, turn it on again
+      if (mapView.locationDisplay.autoPanMode == LocationDisplay.AutoPanMode.OFF) {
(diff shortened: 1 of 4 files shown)


kotlin/show-location-history/src/main/java/com/esri/arcgisruntime/sample/showlocationhistory/MainActivity.kt
```diff
@@ -102,8 +102,14 @@ class MainActivity : AppCompatActivity() {
...
     }
 
-    // change the isTrackLocation flag and the button's icon
+    // reset location display and change the isTrackLocation flag and the button's icon
     button.setOnClickListener {
+      // if the user has panned away from the location display, turn it on again
+      if (mapView.locationDisplay.autoPanMode == LocationDisplay.AutoPanMode.OFF) {
+        mapView.locationDisplay.autoPanMode = LocationDisplay.AutoPanMode.RECENTER
+        // don't change the isTrackLocation status if the button is recentering
+        return@setOnClickListener
```
TADraeseke, line 111: I actually wonder if this is actually more confusing?

It's more intuitive to me that the toggle always toggles on tap and will re-center if applicable. So, I'd suggest dropping the return. Otherwise, looks good!

hud10837: Yeah, that makes sense to me. The only other option would be to have two buttons, which I don't think is necessary here

hud10837:
@TADraeseke Thanks, made those changes!

hud10837:
Thanks! Merging now

PR #849, opened 2020-08-26
add try/catch to configure subnetwork trace add

hud10837:
@TADraeseke I added a try/catch to fix the issue with no value entered in the add condition method. There may be a smarter way to ensure the value is correct, but this seemed like it would catch all possible issues. Could you take a look?


TADraeseke (approved):
LGTM! Thanks @hud10837!

PR #858, opened 2020-10-19
update edit with branch versioning

TADraeseke:
Hey @hud10837 - Time for a bit of a blast from the past? Thanks!


hud10837 (approved):
@TADraeseke These changes look great, super easy to read! I like that the edit dialog appears after changing the location. Everything looks good to me!

PR #862, opened 2020-11-26
Samples to new metered basemaps

TADraeseke:
This PR contains changes to add API keys to samples, but only those samples that use the new BasemapStyles.

It also included some code cleanup


hud10837 (approved):
Looks good to me! Only a few tiny comments but nothing breaking. Well done, that's a huge amount of work!

PR #873, opened 2021-03-09
fix Blurry Brest samples 

TADraeseke:
Fixes for distance measurement analysis, scene layer selection, viewshed camera and scene layer samples.


java/viewshed-camera/src/main/res/values/strings.xml
```diff
@@ -1,8 +1,4 @@
 <resources>
     <string name="app_name">Viewshed Camera</string>
     <string name="update_from_camera">Update from camera</string>
-    <string name="brest_dtm">https://scene.arcgis.com/arcgis/rest/services/BREST_DTM_1M/ImageServer</string>
-    <string name="brest_building_layer">
-        https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0
```
Rachael-E, line 6: my only thought here is wondering why the new Girona string isn't going into the resources folder as the Brest one was previously (it's now directly in the main java class)? I don't think it really matters, but just curious if it was Android/Android samples specific pattern to leave them in here then we could do that.

java/viewshed-camera/src/main/java/com/esri/arcgisruntime/sample/viewshedcamera/MainActivity.java
```diff
@@ -48,18 +49,20 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
     // add base surface for elevation data to the scene view
     Surface surface = new Surface();
-    ArcGISTiledElevationSource brestElevationSource = new ArcGISTiledElevationSource(getString(R.string.brest_dtm));
+    ArcGISTiledElevationSource brestElevationSource = new ArcGISTiledElevationSource(
+        "https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer");
     surface.getElevationSources().add(brestElevationSource);
     scene.setBaseSurface(surface);
 
     // add a scene layer to the scene
```
Rachael-E, line 62:
```suggestion
    // add an integrated mesh layer to the scene
```

Rachael-E (approved):
just a minor comment update and a question about the resources file: otherwise looks good, thanks @TADraeseke !

java/viewshed-camera/src/main/java/com/esri/arcgisruntime/sample/viewshedcamera/MainActivity.java
```diff
@@ -48,18 +49,20 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
     // set the camera to the scene view
-    Camera camera = new Camera(48.3808, -4.49492, 48.2511, 344.488, 74.1212, 0.0);
+    Camera camera = new Camera(41.985, 2.82691, 124.987, 332.131, 82.4732, 0.0);
     mSceneView.setViewpointCamera(camera);
 
     // add base surface for elevation data to the scene view
     Surface surface = new Surface();
-    ArcGISTiledElevationSource brestElevationSource = new ArcGISTiledElevationSource(getString(R.string.brest_dtm));
+    ArcGISTiledElevationSource brestElevationSource = new ArcGISTiledElevationSource(
```
Rachael-E, line 57: also this could be renamed just to `elevationSource`

java/distance-measurement-analysis/src/main/java/com/esri/arcgisruntime/sample/distancemeasurementanalysis/MainActivity.java
```diff
@@ -64,14 +65,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
     // inflate views from layout
     mSceneView = findViewById(R.id.sceneView);
     mDirectDistance = findViewById(R.id.totalDistance);
     mHorizontalDistance = findViewById(R.id.horizontalDistance);
     mVerticalDistance = findViewById(R.id.verticalDistance);
-    Spinner mUnitSpinner = findViewById(R.id.units_spinner);
+    Spinner unitSpinner = findViewById(R.id.units_spinner);
     // set the scene to the view
     mSceneView.setScene(scene);
```
Rachael-E, line 78: to avoid the scene layer cutting into the terrain, we're using an altitude offset on the scene layer:

``` java
    sceneLayer.setAltitudeOffset(1); // offset for visual purposes
```

java/viewshed-camera/src/main/res/values/strings.xml
TADraeseke: So Android Studio linter used to complain about all strings in code not being in the `strings.xml` file. This was to do with localization (where translators would theoretically provide alternatives for different spoken languages). That's not a requirement for URLs, so where possible I'm trying to migrate URLs from `strings.xml` into the code directly (as it's also more readable that way)

TADraeseke:
Thanks @Rachael-E - Made those fixes!

PR #882, opened 2021-05-11
Trev8939/api key basemaps

TADraeseke:
https://devtopia.esri.com/runtime/common-samples/issues/2683

### DIFF
--- a/java/create-and-save-map/README.md
+++ b/java/create-and-save-map/README.md
@@ -25,7 +25,7 @@ Use the navigation drawer to select the basemap and layers you'd like to add to 
 
 ## Additional information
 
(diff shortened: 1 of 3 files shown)


java/create-and-save-map/README.md
```diff
@@ -23,6 +23,10 @@ Use the navigation drawer to select the basemap and layers you'd like to add to
 * ArcGISMap
 * Portal
 
+## Additional information
+
+In this sample, API keys are set directly on the `Basemap` rather than on the `ArcGISRuntimeEnvironment`. This is done to allow the developer to provide the app with access to a range of `BasemapStyle`s, but also allow the user to save the final map to their own ArcGISOnline account.
```
Rachael-E, line 28: ah this PR must have gone in before we agreed on the final wording in the common samples repo: here we go! 

```suggestion
In this sample, an API key is set directly on `Basemap` objects rather than on the whole app using the `ArcGISRuntimeEnvironment` class. This is useful in a scenario where an individual developer is part of an organisation within ArcGIS Online that uses an API key to access a range of `BasemapStyle`s. In the case that an individual member of the organisation wants to save a map locally to their account, and not that of the organisation, they can set the organization's API key on the basemap, and log in to their own account when challenged.  The individual can then save the final map to their own ArcGIS Online account.
```

Rachael-E (approved):
thanks @TADraeseke looks good. just need to update the readme to the wording we agreed on in the samples repo and then this is good to go!

java/create-and-save-map/README.md
TADraeseke: Ooo thanks!

shubham7109:
Looks good to me!

PR #906, opened 2021-08-12
add intent flags and URI handling to fix edit feature attachments

TADraeseke:
Another bug fix for review when you can @shubham7109!

This fixes a case where when attempting to view an attachment it would not start the gallery view due to Uri's going beyond permission scope of the app.


shubham7109 (commented):
Looks good!

shubham7109 (approved):

PR #916, opened 2021-08-30
Update API version mentioned in root README

mikewilburn:
This PR just touches the top-level README to catch one change which was missed with the recent 100.12 release.


shubham7109 (approved):
Thanks for the update

PR #923, opened 2021-10-20
Update samples from named user to API keys 

shubham7109:
This PR updates named user authentication workflow and replace with the preferred API key authentication workflow instead. It also updates existing sample code where appropriate and READMEs.

- [x] Generate offline map (Java & Kotlin)
- [x] Generate offline map override (Java)
- [x] Generate offline map with local basemap (Java)
- [x] Navigate in AR (Java)


shubham7109:
- Removed write permissions Generate Offline Map Overrides
- Reverted code in Create and save map. 

Ready for your review again @TADraeseke

PR #929, opened 2021-11-26
initialize address finding only once map has loaded

gunt0001:
Initializing the address finding UI (spinner control) in the activity's onCreate() can cause the viewpoint of the MapView to be set to the default address extent before the MapView is initialized (before map has loaded). These changes address this issue by intializing the address finding UI only once the map has loaded (see `initializeAddressFinding()`). 

In addition:
- the initial viewpoint is set on the appropriate map property (ArcGISMap.initialViewpoint) instead of setting it on the MapView directly
- removed redundant call to `createPinSymbol()`


fpearson1111:
I also added some commits:
- prevent the callout from showing until the async viewpoint setting operation completes -- seems like showing the callout cancels the viewpoint setting operation
- don't show the callout if it's already being shown -- this stops the callout animation from playing when the callout is already active, which looks "unnatural"

kotlin/find-address/src/main/java/com/esri/arcgisruntime/sample/findaddress/MainActivity.kt
```diff
@@ -85,12 +85,22 @@ class MainActivity : AppCompatActivity() {
...
     val topographicMap = ArcGISMap(BasemapStyle.ARCGIS_STREETS)
+    // set the map viewpoint to start over North America
+    topographicMap.initialViewpoint = Viewpoint(40.0, -100.0, 100000000.0)
+
+    // once the map has loaded successfully, set up address finding UI
+    topographicMap.addDoneLoadingListener {
+      if (topographicMap.loadStatus == LoadStatus.LOADED) {
+        initializeAddressFinding()
+      } else {
+        Toast.makeText(applicationContext, "Map failed to load", Toast.LENGTH_LONG).show()
```
TADraeseke, line 96:
```suggestion
                Log.e(TAG, "Map failed to load: " + topographicMap.loadError.message)
                Toast.makeText(
                    applicationContext,
                    "Map failed to load: " + topographicMap.loadError.message,
                    Toast.LENGTH_LONG
                ).show()
```

TADraeseke (changes requested):
Thanks guys! Just one small suggestion.

We should also change the Java version of this sample if we're changing the behaviour in the Kotlin sample.

fpearson1111:
Thanks @TADraeseke, I made that change, can you take another look please?

shubham7109 (approved):
I've added the Java sample changes, and the Kotlin sample looks good! Could you take a look at this? @TADraeseke

saratchandrakarumuri (approved):
Tested both Java and Kotlin samples and correct viewpoints are set for search results as expected with this fix.

TADraeseke (approved):
LGTM! Thanks @shubham7109

PR #931, opened 2021-12-09
Fix for U13

shubham7109:
Removed `apply plugin: 'kotlin-kapt'` to work with `arcgisVersion=13` as defined in the `version.gradle` file. 

The show-popup sample was built on version 12, where the toolkit required the sample to use the plugin `apply plugin: 'kotlin-kapt'`. The version upgrade [here](https://github.com/Esri/arcgis-runtime-samples-android/commit/74d900d3e740cce487711898146296928beabf48) no longer the needs the plugin which may have caused the build failures.


sorenoid (approved):
LGTM

PR #936, opened 2021-12-14
New sample: Browse Building Floors

shubham7109:
PR do add `Browse building floors` sample to the repo. This sample displays a floor-aware web map and can browse through the building floors.


shubham7109:
Seems good to Trevor, added @TADraeseke's changes from #938.

PR #938, opened 2021-12-15
Trev8939/browse building floors suggestions

TADraeseke:
Nice sample @shubham7109 -- just a few things captured below that I only put in a PR, rather than as suggestions, so we can sneak it into the release.

They're just suggestions, so do check if you agree!


shubham7109:
The changes looks good and really helpful. I'm going to keep this line since we need this to keep the spinner position and the level synced up if the device rotates and the `MapView` rebuilds. 
```kotlin
        // update the spinner to the currently selected floor
        levelSpinner.setSelection(currentFloor)
```

PR #944, opened 2022-01-11
Update script to use Viewbinding

shubham7109:
- Updated the new module script to use Viewbinding for the `MainAcitivty.kt` template
- Updated the `.jar` file 
- Tested and works with the same [run commands](https://github.com/Esri/arcgis-runtime-samples-android/tree/main/tools).


TADraeseke (approved):
LGTM! Good to be using this pattern going forward :D

PR #950, opened 2022-02-01
Trev8939/basemap styles all

TADraeseke:
PR to update the last of our old basemaps to the new basemapstyles

PR #952, opened 2022-02-09
Update samples use basemaps style  

shubham7109:
Same as #950 but need to merge into `main` branch


TADraeseke:
Thanks @shubham7109!

PR #954, opened 2022-02-12
Remove ArcGIS vector tiled layer url sample

shubham7109:
PR to remove the Java sample "ArcGIS vector tiled layer url sample" since we should be promoting the latest recommended workflow. In this case, we will keep the "Vector tiled layer (custom style)" sample.


TADraeseke (approved):
LGTM! Thanks @shubham7109! When building the sample viewer locally after this change, we'll need to remember to delete the assets->codeFiles and assets->metadata folders to make sure none of this old sample is still hanging around

shubham7109:
That's a good point @TADraeseke. I have gone into the habit of clearing the files within "codeFiles" and "assets" as they regularly change with more updated info.

PR #957, opened 2022-02-24
Trev8939/style wms url update

TADraeseke:
This updates the Android sample to use the correct layer for this data set. All other platforms are using this layer.

It will likely break image comparison tests @saratchandrakarumuri.


shubham7109:
Looks good, thanks @TADraeseke  👍

But I would refrain from merging this PR to `main` since this branch contains new sample changes that would need to remain on `v.next` until it is ready for release. Although being so close to a release, this wouldn't really make too much of a difference, but it would be good to follow the new release cycle plan.

PR #958, opened 2022-02-24
Update style WMS layer (URL)

shubham7109:
From @TADraeseke:

> This updates the Android sample to use the correct layer for this data set. All other platforms are using this layer. It will likely break image comparison tests @saratchandrakarumuri.


saratchandrakarumuri (approved):
Changes looking good

PR #959, opened 2022-02-25
Nearest vertex patch

shubham7109:
## Description
PR to update the sample `Nearest Vertex`  to use a distance-preserving projected coordinate system to work with the current planar nearest APIs.

## Links and Data
Devtopia issue: [Link](https://devtopia.esri.com/runtime/common-samples/issues/2997)

## What To Review

- Updated the code to use the California zone 5 (ftUS) state plane coordinate system as the spatial reference.
- Changed the distance measurement to use miles to be consistent with other platforms. 
- Matched the portal item to use the USA States (Generalized) portal item. 

## How to Test
The sample can be cross-verified to see if the distance result values are the same as seen on another source.


kotlin/nearest-vertex/src/main/java/com/esri/arcgisruntime/sample/nearestvertex/MainActivity.kt
```diff
@@ -112,8 +113,13 @@ class MainActivity : AppCompatActivity() {
...
     mapView.apply {
-      // create a map with a basemap and add it to the map view
-      map = ArcGISMap(BasemapStyle.ARCGIS_TOPOGRAPHIC)
+      // create a map using the PortalItem
+      // and add the FeatureLayer to the map view's basemap
+      map = ArcGISMap(statePlaneCaliforniaZone5)
+
+      val portalItem = PortalItem(Portal("https://arcgisruntime.maps.arcgis.com",false),"99fd67933e754a1181cc755146be21ca")
+      val usStatesGeneralizedLayer = FeatureLayer(portalItem,0)
+      map.basemap.baseLayers.add(usStatesGeneralizedLayer)
```
TADraeseke, line 122: needs formatting

kotlin/nearest-vertex/src/main/java/com/esri/arcgisruntime/sample/nearestvertex/MainActivity.kt
```diff
@@ -63,6 +61,9 @@ class MainActivity : AppCompatActivity() {
     activityMainBinding.coordinateDistanceTextView
   }
 
+  // California zone 5 (ftUS) state plane coordinate system.
+  private val statePlaneCaliforniaZone5 = SpatialReference.create(2229)
```
TADraeseke, line 65:
```suggestion
  private val statePlaneCaliforniaZone5SpatialReference = SpatialReference.create(2229)
```

Just think it'll be easier to read later that the PointCollection(...) is taking a spatial reference as an argument

TADraeseke (approved):
Thanks @shubham7109 -- Just a couple small changes. Looks good!

PR #970, opened 2022-04-13
Map image layer fix

shubham7109:
## Description
PR to a fix for the java sample `ArcGIS Map Image Layer (URL)`. The map was not rendering on the sample viewer due to a  conflict in the name of the string used. 

## Links and Data
Sample Epic: `runtime/common-samples/issues/3239`


TADraeseke (approved):
LGTM!

PR #971, opened 2022-04-15
Trev8939/basemap style final check

TADraeseke:
Fix the last of the Basemap.Type code and comments lingering


shubham7109 (approved):
Changes look good and tested on the sample viewer 👍

PR #972, opened 2022-04-15
Sample Patches

shubham7109:
Resolved the following issues found during full-cert

- [x]  Add features - Fixed the use of `applyEditsAsync()`
- [x] Add contingent values- Added statement to check if file exists to avoid crash
- [x] Delete features - Fixed the use of `applyEditsAsync()`


saratchandrakarumuri (approved):
Changes are looking good... Issues with all three samples are fixed with this

PR #974, opened 2022-04-18
Create and Save KML point symbols patch

shubham7109:
## Description
`Create and Save KML` sample has an issue with the "Point Symbol" images not loading using a spinner adapter. Fixed by using Viewbinding on `getView()` instead of `createView()` 

## Links and Data
Sample Epic: `runtime/common-samples/issues/3452`


saratchandrakarumuri (approved):
All point symbols are now rendered as expected with this PR

TADraeseke (changes requested):
Great to get this fixes and use GONE instead of invisible. UI looks cleaner.

The only thing is the bottom box now partly covers the attribution bar. I think either bump the UI box up or put it back in the upper right.

Thanks @shubham7109!

shubham7109:
Bumped up the UI box to not block the attribution bar

TADraeseke (approved):
LGTM! Thanks @shubham7109!

PR #975, opened 2022-04-19
add API key to point scene layer sample

TADraeseke:
Fix for white rendering of Add point scene layer sample


saratchandrakarumuri (approved):
Fix is working as expected

PR #979, opened 2022-04-28
New module script update

shubham7109:
## Description
PR to update the template file and rebuild the output JAR file to use TAG. We use the TAG values to help log errors which is useful for debugging purposes.

Added the line to the new module script `private val TAG: String = MainActivity::class.java.simpleName`.

## Links and Data
Sample Epic: `runtime/common-samples/issues/3466`


TADraeseke (approved):
LGTM!

PR #998, opened 2022-07-20
Geotriggers sample patch

shubham7109:
## Description
PR to rename the Geotrigger's sample file name from `ListItemBinding` to `GardenListItemBinding`. The original file name was causing a file name conflict in the sample viewer causing the sample to crash.


TADraeseke (approved):
Thanks for catching this fix @shubham7109!

PR #1001, opened 2022-07-23
Test PR

shubham7109:
This is to test the metadata script. Will delete later


shubham7109:
@TADraeseke 
I added a new sample to test the script and check for various edge cases. Most situations the test handled the malformed data well. But with all the changes for a new sample in place, the script still fails with the message

>   Error parsing paths - display-test-map - Unable to get kotlin or java source code paths..
Error populate failed for - display-test-map.
Unable to get kotlin or java source code paths.

TADraeseke:
The final check not working because of a missing `redirect_from` field in the metadata. The other new PR does have a blank `redirect_from` field (and passes checks). Which do we prefer to be the norm going forward? A blank `redirect_from` by default? Or no `redirect_from` by default? I guess my preference is a blank one for easy editing in the future, but curious your thoughts :)

https://github.com/Esri/arcgis-runtime-samples-android/blob/fce85321393ee2965cf235418631d52319fade5a/kotlin/render-multilayer-symbols/README.metadata.json#L26

TADraeseke:
@shubham7109 the script now takes care of provisioning to and from. So I'm going to mark this closed for now and we can improve the script as we go :)

PR #1007, opened 2022-08-04
update cql codes

TADraeseke:
## Description

Add new CQL filters

NOTE: I had to change the Theme to prevent a crash. Don't know if this might effect sample viewer


shubham7109:
CQL queries now work! The theme change prevented the crash but was not loading the data/time material views. I made changes to support the material theme for the sample by removing the `AppCompat` dependency.

shubham7109 (approved):

saratchandrakarumuri (approved):
`AP010` query is no longer crashing the app and is working good as expected

